PR #2, opened 2022-12-19
Create simple GitHub actions CI

Flova:


Juphex:
the yml file states "ubuntu-latest".
your readme states Ubunutu 20.04

.github/workflows/main.yml
```diff
@@ -0,0 +1,13 @@
+name: CI
+
+on: [pull_request, workflow_dispatch]
```
Juphex, line 3: I would also suggest to add push on main.
so maybe add:

`push:
         branches: [main]`

Flova:
ROS 1 support officially ends with Ubuntu 20.04. Afterwards, one needs to build everything from source and apply some patches for changed library versions. You can talk to Michael about this as he is trying to maintain ROS 1 past 20.04 (aka. ROS noetic). We are therefore limited to 20.04 :/

As it seems, it is the case that the ROS Industrial pipeline builds on top of docker inside the github actions container. The action container therefore uses the latest Ubuntu, but the docker container that is used to build and test it uses 20.04.

.github/workflows/main.yml
Flova: Due to our settings the main can not be modified without a pull request.

Juphex: How does pull against master behave?
Is it seen like a push to push when merged?

.github/workflows/main.yml
```diff
@@ -0,0 +1,13 @@
...
+
+jobs:
+  main_ci:
+    runs-on: ubuntu-latest
```
Juphex, line 7: ubuntu-20.04

Flova: The industrial action runs the docker instance etc. on 22.04 aka latest as described in the documentation. Only the container that is started inside runs 20.04. This is kind of weird I know, but they do it this way. And we have no benefit of running the host system at an older Ubuntu Version.

Juphex (approved):

.github/workflows/main.yml
Flova: If I understand you correctly, then yes.

PR #3, opened 2022-12-19
Fix workspace creation in the the README.md

Flova:


README.md
```diff
@@ -27,7 +27,8 @@ mkdir -p catkin_ws/src
 
 # Clone repository into source folder
 cd catkin_ws/src
-git clone git@github.com:UHHRobotics22-23/robot_project 
+git clone git@github.com:UHHRobotics22-23/robot_project
+cd ..
```
Juphex, line 31: is this correct on linux?
I am used to cd robot_project

Flova: We don't want to go inside the repo, but back to the workspace. In Linux `..` goes one directory up.

Juphex (approved):

PR #5, opened 2022-12-28
Artificial data generation and vision transformer (first minimal prototype)

Flova:
This PR adds a simple script for lilypond based data generation. The script does not include pauses etc. for now and needs to be extended in the future.

I also add a training script for a vision transformer model based on [Donut](https://github.com/clovaai/donut). The model is able to predict the notes on a validation dataset. I tested this with sheets, including a couple hundred notes per sheet (it is a bit overkill, so I did not include it here).


Flova:
Training using the latest version of the `train.py` script is still converging, so I would wait until it finished before merging this PR.

But it is looking good so far. Edit distance on the validation set:
val_metric(1)

Juphex:
could you add versions to your requirements.txt?

Juphex:
would you prefer to stick to a global .gitignore or at package level for this project?

marimbabot_vision/scripts/train.py
```diff
@@ -0,0 +1,221 @@
...
+        for i, results in enumerate(validation_step_outputs):
+            for scores in results:
+                cnt[i] += len(scores)
+                total_metric[i] += np.sum(scores)
+            val_metric[i] = total_metric[i] / cnt[i]
+            val_metric_name = f"val_metric_{i}th_dataset"
+            self.log_dict({val_metric_name: val_metric[i]}, sync_dist=True)
+        self.log_dict({"val_metric": np.sum(total_metric) / np.sum(cnt)}, sync_dist=True)
+
+    def configure_optimizers(self):
```
Juphex, line 193: why is it called optimizers?

marimbabot_vision/scripts/train.py
```diff
@@ -0,0 +1,221 @@
...
+
+# Create pytorch lightning module
+class DonutModelPLModule(pl.LightningModule):
+    def __init__(self, config, pre_processor, model):
+        super().__init__()
+        self.config = config
+        self.pre_processor = pre_processor
+        self.model = model
+
+    def training_step(self, batch, batch_idx):
```
Juphex, line 131: a "step" equals one batch here?

marimbabot_vision/scripts/train.py
```diff
@@ -0,0 +1,221 @@
...
+            val_metric[i] = total_metric[i] / cnt[i]
+            val_metric_name = f"val_metric_{i}th_dataset"
+            self.log_dict({val_metric_name: val_metric[i]}, sync_dist=True)
+        self.log_dict({"val_metric": np.sum(total_metric) / np.sum(cnt)}, sync_dist=True)
+
+    def configure_optimizers(self):
+        return torch.optim.Adam(self.parameters(), lr=self.config.get("lr"))
+
+    def train_dataloader(self):
+        return train_dataloader
```
Juphex, line 197: would make more sense using an attribute for the data loader?
=> return self.train_dataloader

Juphex (changes requested):
could you add versions to your requirements.txt?

marimbabot_vision/scripts/train.py
Flova: Following the interface define here: https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html

marimbabot_vision/scripts/train.py
Flova: Yes, we process a single batch each step. 

Also see: https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html

marimbabot_vision/scripts/train.py
Flova: We could do this. Now the reference is injected in an ugly way and your way would be a bit more verbose but more clean in the long run if we split the script in multiple files.

Juphex (approved):
lgtm
I would still prefer to have a requirements.txt provided with versions

requirements.txt
```diff
@@ -0,0 +1,8 @@
+abjad
+numpy
+tqdm
+nltk
+pillow
+pytorch-lightning
+torch
+transformers
```
Juphex, line 8: versions?

Flova: I was only able to train this with one machine running Arch and one running Ubuntu 22.04. As we are targeting ROS1 with Ubuntu 20.04 the python major version differs and pinning the versions to my version might be risky and lead to some packages not working or unnecessary compilation from source by pip. Overall pining versions is important, but I would be careful doing it blindly.

Flova:
I also added some fixes in the sampling that otherwise resulted in broken time signatures.

requirements.txt
Juphex: alright let's do that in a future commit

Juphex (approved):

PR #6, opened 2023-01-16
add Python version / package versions?

Juphex:
I would like to add the Python version / package versions. It seems that some packages only work on Python 3.10. I added that in the docs. I guess it would make sense to add package versions as well, @Flova  could you help me out or even add them here when appropriate?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,4 @@ Pillow==9.4.0
 pytorch_lightning==1.8.1
 torch==1.13.1
 tqdm==4.64.0
-transformers==4.25.1
-sentencepiece
+transformers[sentencepiece]==4.25.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-abjad==3.15
+abjad==3.4
 nltk==3.7
 numpy==1.23.3
 Pillow==9.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
-abjad
-numpy
-tqdm
-nltk
-pillow
-pytorch-lightning
-torch
-transformers
+abjad==3.15
+nltk==3.7
+numpy==1.23.3
+Pillow==9.4.0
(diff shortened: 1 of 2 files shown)


Flova:
We use ros noetic, which is based on Ubuntu 20.04. 
The targeted python version for Ubuntu 20.04 is python 3.8 as stated here: https://wiki.ubuntu.com/FocalFossa/ReleaseNotes

I would strongly prefer using the base python version of the Ubuntu release as it makes makes our life way easier (e.g. on lab machines where we have no root access. There are ways around it but we want to keep the setup easier for people that are not that into python and ros.

That being said I currently don't have a Ubuntu 20.04 machine that is capable of training my vision code, so I would prefer targeting a wider range for that.

We can also use something like poetry to calculate our dependency graph for a range of python versions and lock the packages, so the setup is the same everywhere. But it is a bit more complex as one would need to install poetry, install the deps with it and source also the poetry environment.

Flova:
Which ones where only compatible with 3.10?

Juphex:
abjad==3.15
@Flova

requirements.txt
```diff
@@ -1,8 +1,9 @@
...
\ No newline at end of file
+abjad==3.4
+nltk==3.7
+numpy==1.23.3
+Pillow==9.4.0
+pytorch_lightning==1.8.1
+torch==1.13.1
+tqdm==4.64.0
+transformers==4.25.1
+sentencepiece
```
Flova, line 9: Why is there no version for sentencepiece?

Flova:
Also #8 adds new deps, so we might want to wait for them and bump this one. Or we merge this one now and add versions in the other PR.

requirements.txt
Juphex: added it to the transformers line now

Juphex:
if you are fine with sentencepiece lets merge this @Flova

Flova (approved):

PR #7, opened 2023-01-16
remove lilypond footer line

Juphex:
remove footer in data generation

### DIFF
--- a/marimbabot_vision/scripts/generate_data.py
+++ b/marimbabot_vision/scripts/generate_data.py
@@ -44,8 +44,11 @@ def generate_sample(i):
     """Generate a sample and save it to disk"""
     string, staff = generate_piece()
     os.makedirs(f"{OUTPUT_DIR}/{i}", exist_ok=True)
+    header_block = Block(name="header")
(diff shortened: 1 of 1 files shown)


marimbabot_vision/scripts/generate_data.py
```diff
@@ -44,8 +44,11 @@ def generate_sample(i):
     """Generate a sample and save it to disk"""
     string, staff = generate_piece()
     os.makedirs(f"{OUTPUT_DIR}/{i}", exist_ok=True)
+    header_block = Block(name="header")
+    header_block.tagline = "#ff"
```
Flova, line 48: Is it possible to set the tag line in the same way you set the name?

Juphex: I don't know, this is how they proposed it in the docs.

Juphex: Yes. Before and after below.
big-staff_1
big-staff_1

Flova (approved):

PR #8, opened 2023-01-18
Feature/image augmentation

Juphex:


Juphex:
flipping not (yet) included

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -26,16 +26,17 @@ def augment_sample(i):
...
-        "random_contrast": [A.RandomBrightnessContrast(always_apply=True)],
         "blur": [A.Blur(blur_limit=15, always_apply=True)],
-        "median_blur" : [A.MedianBlur(blur_limit=15, always_apply=True)],
-        "glass_blur": [A.GlassBlur(always_apply=True)],
+        "median_blur" : [A.MedianBlur(blur_limit=5, always_apply=True)],
+        "glass_blur": [A.GlassBlur(sigma=0.4,always_apply=True)],
         "zoom_blur": [A.ZoomBlur(always_apply=True)],
-        "image_compression": [A.ImageCompression(quality_lower=20 ,always_apply=True)],
-        "downscale": [A.Downscale(scale_min=0.5, scale_max=0.5, always_apply=True, interpolation=cv2.INTER_NEAREST)],
+        "image_compression": [A.ImageCompression(quality_lower=60, quality_upper=60 ,always_apply=True)],
```
Flova, line 33: Why do we always apply compression? Isn't this kind of predictable?

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -26,16 +26,17 @@ def augment_sample(i):
...
+        "downscale": [A.Downscale(scale_min=0.4, scale_max=0.4, always_apply=True, interpolation=cv2.INTER_NEAREST)],
```
Flova, line 34: Sane here. Why are min and max the same?

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -26,16 +26,17 @@ def augment_sample(i):
...
         "rotate_3": [A.Rotate(limit=[3,3], always_apply=True)],
-        "rotate_6": [A.Rotate(limit=[6,6], always_apply=True)]
+        "rotate_6": [A.Rotate(limit=[6,6], always_apply=True)],
+        "translate_x_16": [A.Affine(translate_px={"x":16})],
+        "translate_y_16": [A.Affine(translate_px={"y":16})],
```
Flova, line 38: Why do we do 3 separate affine transformations? They should be combinable.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,52 @@
...
+from multiprocessing import Pool
+from PIL import Image
+
+import tqdm
+import shutil
+import albumentations as A
+import numpy as np
+import cv2
+
+def apply_transforms(img, transforms = [A.RGBShift()]):
```
Flova, line 11: Instantiating objects in a conditional argument is a bad practice in python.

Flova: It does not really matter here but they are only initialized once and this might lead to weird behavior.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,52 @@
...
+        "image_compression": [A.ImageCompression(quality_lower=60, quality_upper=60 ,always_apply=True)],
+        "downscale": [A.Downscale(scale_min=0.4, scale_max=0.4, always_apply=True, interpolation=cv2.INTER_NEAREST)],
+        "rotate_3": [A.Rotate(limit=[3,3], always_apply=True)],
+        "rotate_6": [A.Rotate(limit=[6,6], always_apply=True)],
+        "translate_x_16": [A.Affine(translate_px={"x":16})],
+        "translate_y_16": [A.Affine(translate_px={"y":16})],
+        "translate_x_y_15": [A.Affine(translate_px={"y":15, "x":15})],
+        })
+
+    for key in transformations.keys():
```
Flova, line 42: Instead of applying transformations individually, I would strongly prefer creating a composition of all augmentations (with `always_apply=False`).

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,52 @@
...
+        new_img = apply_transforms(img, transforms=transformations[key])
+
+        Image.fromarray(new_img).save(orig_img_path.replace(".png", "") + f"_{key}.png")
+        shutil.copy(orig_txt_path, orig_txt_path.replace(".txt", "") + f"_{key}.txt")
+        shutil.copy(orig_ly_path, orig_ly_path.replace(".ly", "") + f"_{key}.ly")
```
Flova, line 47: To keep the dataset definition consistent and enable easier loading, I would like to create only one augmented version for each sequence. Generating more sequences is cheap (at least in simulation) and we don't want to overfit on them. So having a new sequence for each randomly sampled augmentation is the way to go imo.
I would also prefer placing the augmented samples in a completely different directory, so that the format for each sample is consistent, and we can just change the path of the dataset for the training.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,52 @@
...
+
+if __name__ == "__main__":
+    # Call generate_sample on ids with tqdm and multiprocessing (lilypond is single threaded)
```
Flova, line 50: We don't use lilypond here :P

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,52 @@
+from generate_data import NUM_SAMPLES, NUM_WORKER, OUTPUT_DIR
```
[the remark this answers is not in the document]
Flova: Or keep the variables locally.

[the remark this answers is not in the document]
Flova: Or more ideally we would use [argparse](https://docs.python.org/3/library/argparse.html) and cli arguments with sane default values.

Flova (changes requested):

marimbabot_vision/scripts/data_augmentation.py
Juphex: you are right

marimbabot_vision/scripts/data_augmentation.py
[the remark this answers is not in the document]
Juphex: I will do this in another PR. Or did you do it already?

Juphex:
I added a different output directory for the data.
I combined transformations and only create 1 example by now. @Flova

marimbabot_vision/scripts/data_augmentation.py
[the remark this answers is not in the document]
Flova: No I did not do it

.gitignore
```diff
@@ -1,6 +1,7 @@
 *.pyc
 __pycache__/
 marimbabot_vision/scripts/data/
+marimbabot_vision/scripts/data_augmented/
```
Flova, line 4: The upstream gut ignore features a `data_*` rule iirc. So this would be redundant

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,50 @@
...
+from multiprocessing import Pool
+from PIL import Image
+
+import os
+import tqdm
+import shutil
+import albumentations as A
```
Flova, line 8: Not a mayor issue, but can you sort the imports? In VSCode it is just right click -> sort imports

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,50 @@
...
+    transformed_image = transformed["image"]
+
+    return transformed_image
+
+def augment_sample(i, amount = 1):
+    """Generate a augmentations for a given sample and save it to disk"""
+    orig_img_path = f"{OUTPUT_DIR}/{i}/staff_1.png"
+    orig_txt_path = f"{OUTPUT_DIR}/{i}/staff_1.txt"
+    orig_ly_path = f"{OUTPUT_DIR}/{i}/staff_1.ly"
+    img = Image.open(orig_img_path)
```
Flova, line 28: You can use opencv's imread here. So you don't need to cast the image into a NumPy array later on.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,50 @@
...
+
+    # set of simpler transformations
+    transformations = [
```
Flova, line 31: The transformations can be defined once at the top of the file so they are not instanciated every time and it easier to see what is applied.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,50 @@
...
+    transformations = [ 
+        A.MedianBlur(blur_limit=5,), A.ZoomBlur(max_factor=1.1), \
+        A.Downscale(scale_min=0.6, scale_max=0.99, interpolation=cv2.INTER_NEAREST), \
+        A.Affine(translate_px={"y":10, "x":10}, rotate=[-3,3]) 
+        ]
+    
+    augmented_path = f"{AUGMENT_OUTPUT_DIR}/{i}"
+    os.makedirs(augmented_path, exist_ok=True)
+
+    for i in range(amount):
```
Flova, line 40: The for loop is not nessesary anymore.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,50 @@
...
+        new_img = apply_transforms(img, transformations)
+
+        Image.fromarray(new_img).save(os.path.join(augmented_path, f"staff_1_augment_{i}.png"))
```
Flova, line 43: This is not following the naming we used elsewhere. Especially since we only generate one sample now just call it staff_1.{txt,png,ly}.

marimbabot_vision/scripts/data_augmentation.py
```diff
@@ -0,0 +1,50 @@
...
+        shutil.copy(orig_txt_path, os.path.join(augmented_path, f"staff_1_augment_{i}.txt"))
+        shutil.copy(orig_ly_path, os.path.join(augmented_path, f"staff_1_augment_{i}.ly"))
+
+if __name__ == "__main__":
+    # Call generate_sample on ids with tqdm and multiprocessing
+    with Pool(NUM_WORKER) as pool:
+        list(tqdm.tqdm(pool.imap(augment_sample, range(NUM_SAMPLES)), total=NUM_SAMPLES))
```
Flova, line 50: This assumes that the dataset we augment is named as a consistent integer sequence. Using os.listdir() like in that script for the handwritten notes would be more reliable. Imagine if e.g. a bad sample gets deleted or prefixed.

Flova (changes requested):

marimbabot_vision/scripts/data_augmentation.py
Flova: Also the comment is outdated.

marimbabot_vision/scripts/data_augmentation.py
Juphex: that does not work for me

marimbabot_vision/scripts/data_augmentation.py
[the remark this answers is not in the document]
Juphex: https://github.com/UHHRobotics22-23/robot_project/pull/13

Flova (approved):

PR #9, opened 2023-01-19
add script for whiteboard sample collection

scaredycode:
@Flova


Flova:
> we may should rename it to a more clear description like live_video_data_collection, for example

I agree. Maybe just `real_world_data_collection.py`?

Flova:
Addressing the comments in #10 and #8 I would like to have a consistent output format. This is also not the case here. But I honestly like it in this case, because we have a 1 to many relationship for each sample and it makes it easier to verify the data afterward. But we should think about adding a script that creates individual samples in the common format in the end.

Flova (approved):

PR #10, opened 2023-01-19
Feature/handwritten data

scaredycode:
Script that generates handwritten music notation by selecting randomly from handwritten symbols.
Iterates through specified data set and stores multiple handwritten images in each sample folder.


marimbabot_vision/scripts/generate_hw_sample.py
```diff
@@ -0,0 +1,114 @@
+import sys
```
Juphex, line 1: remove this import as it is not used

Juphex (commented):
I would prefer to have an image file, e.g. staff_1_hw_4.png and a corresponding ground truth file staff_1_hw_4.txt. At least we need to make a call how to do it because we need consistency with  https://github.com/UHHRobotics22-23/robot_project/pull/8 .

I did check some example and they look correct.

scaredycode:
> I would prefer to have an image file, e.g. staff_1_hw_4.png and a corresponding ground truth file staff_1_hw_4.txt. At least we need to make a call how to do it because we need consistency with #8 . #8 creates an image and a corresponding ground truth file to it.

I'm not sure I understand. The script is intended to be used on a previously generated data set using generate_data.py. 
For each previously generated sample, the handwritten images are then stored in the same folder as the remaining corresponding sample files (staff_1.ly, staff_1.txt, staff_1.png). Handwritten samples are generated using the staff_1.txt file and should therefore be equivalent to the staff_1.png notation. The ground truth would then be staff_1.txt (or staff_1.ly), the same as used for the lilypond visualization (staff_1.png).

Flova:
Similar to what I said in #8 I would prefer a mapping of one sample -> one augmented sample (handwritten in this case). The sample should be stored in a new directly so that is easy to use downstream and we don't need to change the data loading. If we do this handwritten samples could also be augmented using the data augmentation in the same way we augment the normal samples. Meaning we output a new sample with a staff_1.{ly, txt, png} in a new folder.

marimbabot_vision/scripts/generate_hw_sample.py
```diff
@@ -0,0 +1,113 @@
...
+    draw_staff(image, 0)
+    draw_symbol(image, f'{HW_SYMBOLS_DIR}/time', (70, 50))
+    return image
+
+def draw_piece(string, outdir, name):
+    sample_im = generate_sample_image()
+    x_pos = 70 + randint(MIN_SYMBOL_DIST, MAX_SYMBOL_DIST) 
+    y_offset = 0
+    duration_counter = 0
+    piece = [(n[0], n.count('\''), int((re.findall(r'\d+', n)[0]))) for n in string.split()]
```
Flova, line 89: We might want to separate the parser as it might be useful elsewhere.

marimbabot_vision/scripts/generate_hw_sample.py
```diff
@@ -0,0 +1,113 @@
...
+            draw_note(sample_im, (x_pos, y_pos+y_offset), is_flipped, duration)
+            
+        x_pos += randint(MIN_SYMBOL_DIST, MAX_SYMBOL_DIST)
+        duration_counter += 1/duration
+    draw_symbol(sample_im, f'{HW_SYMBOLS_DIR}/bar', (x_pos,50+y_offset))
+
+    sample_im.save(f'{outdir}/{HW_NAME_PREFIX}_{name}.png','PNG')
+
+if __name__ == "__main__":
+    for sample_folder in tqdm.tqdm(sorted(os.listdir(WORKING_DIR), key= lambda x: int(x))):
```
Flova, line 108: I don't think we need to sort them here.

Flova, line 108: Also, how fast is this? We might want to parallelize it.

Flova (changes requested):

marimbabot_vision/scripts/generate_hw_sample.py
```diff
@@ -0,0 +1,113 @@
+import os
+import tqdm
+from numpy.random import choice, randint
+from PIL import Image, ImageDraw
+import re
+
+# generates handwritten music notation by selecting randomly from handwritten symbols
+# iterates through WORKING_DIR and stores multiple handwritten images in each sample folder
+
+WORKING_DIR = 'data'
```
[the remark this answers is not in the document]
scaredycode: changed it to INPUT_DIR and added new variable for output directory OUTPUT_DIR

marimbabot_vision/scripts/generate_hw_sample.py
```diff
@@ -0,0 +1,113 @@
...
+
+HW_SYMBOLS_DIR = 'hw_notation'
+HW_NAME_PREFIX = 'staff_1_hw'
```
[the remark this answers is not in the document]
scaredycode: now creates a new directory OUTPUT_DIR where each sample folder is replicated with one handwritten png replacing the lilypond generated png. names for sample folders and files are the same.

marimbabot_vision/scripts/generate_hw_sample.py
```diff
@@ -0,0 +1,113 @@
...
+
+IMAGES_PER_SAMPLE = 5
```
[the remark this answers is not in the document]
scaredycode: removed feature to generate multiple random handwritten samples. Now only generates one handwritten sample per sample folder in given INPUT_DIR.

marimbabot_vision/scripts/generate_hw_sample.py
scaredycode: pretty fast, 100 samples take my laptop approx. 2 seconds.

Flova (approved):
It looks really good. Thank you!

marimbabot_vision/scripts/generate_hw_sample.py
scaredycode: You're right. I probably copied that one from another script ^^''. Removed it, still works.

Juphex (approved):

PR #11, opened 2023-01-21
Add detection script for the transformer model

Flova:
- Add detection script to run a given model on a sample image
- Support model and dataset versions in gitignore
- Fix deprecation warning and slow execution on highly parallel machines (hand written data generation) (a bit off topic sry)

### DIFF
--- a/marimbabot_vision/scripts/detect.py
+++ b/marimbabot_vision/scripts/detect.py
@@ -1,0 +1,58 @@
+import torch
+from PIL import Image
+from transformers import (DonutProcessor, VisionEncoderDecoderConfig,
+                          VisionEncoderDecoderModel)
+
+INPUT_FILE = './demo.png'
+MODEL = './model'
+
+# Load base model
+ved_config = VisionEncoderDecoderConfig.from_pretrained(MODEL)
+pre_processor = DonutProcessor.from_pretrained(MODEL)
+model = VisionEncoderDecoderModel.from_pretrained(MODEL, ignore_mismatched_sizes=True, config=ved_config)
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+model.to(device)
+
+# Load image
+image = Image.open(INPUT_FILE).convert('RGB')
+
+# Rotate image if needed
+if image.size[0] > image.size[1]:
+    image = image.transpose(Image.Transpose.ROTATE_90)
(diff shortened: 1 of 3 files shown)


marimbabot_vision/scripts/generate_hw_data.py
```diff
@@ -108,8 +109,13 @@ def draw_piece(string, sample_name):
...
+    draw_piece(string, sample)
+
+
 if __name__ == "__main__":
-    for n in tqdm.tqdm(os.listdir(INPUT_DIR)):
-        with open(f'{INPUT_DIR}/{n}/staff_1.txt', 'r') as f:
-            string = f.read()
-        draw_piece(string, n)
+
+    with Pool(32) as pool:
```
Flova, line 120: TODO make workers configurable

Juphex: guess it is configurable now?

Flova: Yep

Flova:
Rdy to review @Juphex

marimbabot_vision/scripts/detect.py
```diff
@@ -0,0 +1,58 @@
...
+
+device = "cuda" if torch.cuda.is_available() else "cpu"
+model.to(device)
+
+# Load image
+image = Image.open(INPUT_FILE).convert('RGB')
+
+# Rotate image if needed
+if image.size[0] > image.size[1]:
+    image = image.transpose(Image.Transpose.ROTATE_90)
```
Juphex, line 22: we should figure out a way how to overcome this later because we may not know whether to rotate right or left

Juphex (approved):
I couldn't run the code because I don't have trained wheights here. But it looks good to me

marimbabot_vision/scripts/detect.py
Flova: This is mainly because lilypond generates them in portrait and it writes them directly to disk and I don't want to load and safe them again. And the script from Juli makes them as landscape, which is also reasonable. Currently we only have these two options, so the code works, but it is a bit ugly, I agree with that.

Flova: We can open an issue/ticket for that

PR #14, opened 2023-01-26
Init commit motion planning

berkgungor:
Template for motion planning with moveIt is added and will be modeified with specific position to hit the marimba.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,27 @@
+{
+  "configurations": [
+    {
+      "browse": {
+        "databaseFilename": "${default}",
+        "limitSymbolsToIncludedHeaders": false
+      },
+      "includePath": [
+        "/homeL/project2022/ros/devel/include/**",
+        "/homeL/project2022/diana7_ws/devel/include/**",
+        "/opt/ros/noetic/include/**",
+        "/homeL/project2022/ros/src/bio_ik/include/**",
+        "/homeL/project2022/diana7_ws/src/diana7_ros/diana7_control/include/**",
+        "/homeL/project2022/diana7_ws/src/diana7_ros/diana7_gazebo/include/**",
+        "/homeL/project2022/diana7_ws/src/diana7_ros/diana7_hw/include/**",
+        "/homeL/project2022/diana7_ws/src/diana7_ros/test/include/**",
+        "/usr/include/**"
+      ],
+      "name": "ROS",
+      "intelliSenseMode": "gcc-x64",
+      "compilerPath": "/usr/bin/gcc",
+      "cStandard": "gnu11",
+      "cppStandard": "c++14"
+    }
+  ],
+  "version": 4
+}
--- a/marimbabot_planning/CMakeLists.txt
+++ b/marimbabot_planning/CMakeLists.txt
@@ -1,209 +1,23 @@
 cmake_minimum_required(VERSION 3.0.2)
-project(marimba_move)
+project(marimbabot_planning)
 
-## Compile as C++11, supported in ROS Kinetic and newer
-# add_compile_options(-std=c++11)
-
-## Find catkin macros and libraries
-## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
-## is used, also find other catkin packages
 find_package(catkin REQUIRED COMPONENTS
-  moveit_ros_planning
   moveit_ros_planning_interface
-  moveit_visual_tools
   roscpp
-  std_msgs
   tf2_eigen
+  tf2_ros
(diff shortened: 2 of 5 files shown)


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,27 @@
+{
```
Flova, line 1: Please add the `.vs_code` folder to the gitignore as it contains some machine specific paths.

marimbabot_planning/CMakeLists.txt
```diff
@@ -1,205 +1,23 @@
...
 include_directories(
-# include
   ${catkin_INCLUDE_DIRS}
 )
 
-## Declare a C++ library
-# add_library(${PROJECT_NAME}
-#   src/${PROJECT_NAME}/marimbabot_planning.cpp
-# )
+add_executable(move_to_pose src/move_to_pose.cpp)
```
Flova, line 18: Maybe we want to rename our code to something like `simple_moveit_hit.cpp`.

marimbabot_planning/package.xml
```diff
@@ -8,15 +8,31 @@
   The package is a necessary part of the robot and is required to ensure the robot is able to move accurately and safely.
   </description>
 
-  <maintainer email="root@todo.todo">root</maintainer>
+  <maintainer email="1guengoe@todo.todo">1guengoe</maintainer>
```
Flova, line 11: Small nitpick, the main is todo@todo.

marimbabot_planning/src/move_to_pose.cpp
```diff
@@ -0,0 +1,49 @@
+#include <iostream>
+#include <ros/ros.h>
+#include <moveit/move_group_interface/move_group_interface.h>
+#include <tf2_ros/transform_listener.h>
+#include <geometry_msgs/TransformStamped.h>
+
+
+int main(int argc, char **argv)
+{
+    ros::init(argc, argv, "marimba_move");
```
Flova, line 10: Maybe chose a node name that resamples the file name in a closer way. `marimba_move` is very broad.

marimbabot_planning/src/move_to_pose.cpp
```diff
@@ -0,0 +1,49 @@
...
+    static const std::string PLANNING_GROUP = "arm_with_tcp";
+    moveit::planning_interface::MoveGroupInterface move_group_interface(PLANNING_GROUP);
+
+    const moveit::core::JointModelGroup* joint_model_group =
+        move_group_interface.getCurrentState()->getJointModelGroup(PLANNING_GROUP);
+
+
+    tf2_ros::Buffer tfBuffer;
+    tf2_ros::TransformListener tfListener(tfBuffer);
+    ros::Duration(1).sleep();
```
Flova, line 24: Why do you sleep here? You can configure tf timeouts if this was an issue.

marimbabot_planning/src/move_to_pose.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    geometry_msgs::TransformStamped transformStamped;
+    try{
+        transformStamped = tfBuffer.lookupTransform("world", "diana_gripper/tcp", ros::Time(0));
```
Flova, line 28: Are the frames correct? Is moveit running in the `world` frame?

Juphex (commented):
seems like it is still work in progress?

marimbabot_planning/src/move_to_pose.cpp
berkgungor: without sleep, it got late and threw an error. I am not confident on TF functions so just put duration sleep which works.

marimbabot_planning/src/move_to_pose.cpp
berkgungor: Yes, so far it worked, we got the motion.

berkgungor:
The motion planning code still in template format, will be modified this week.

marimbabot_planning/src/move_to_pose.cpp
v4hn: in principle a sleep after setting up a `TransformListener` is perfectly valid setup procedure.
TF lookup timeouts should represent the amount of time needed to wait during regular operation, and not the startup procedure.
In this concrete case it can be slightly more efficient to set the timeout below because the lookup is done only *once* right after constructing the `TransformListener`. But then again, the much better solution for a single loopup would be to use a [tf2_ros/buffer_server](https://github.com/ros/geometry2/blob/c92cd9a9073c255388599495d2f86799ecafa1bb/tf2_ros/CMakeLists.txt#L54-L63) and a single action.

Flova:
Closed in favor of #26

PR #15, opened 2023-01-30
script for whiteboard data outsourcing

scaredycode:
Takes a data set with multiple whiteboard images per sample (as generated from real_world_data_collection.py) and creates a separate data set in the typical structure with just one image per sample.

Sample names consist of sample number (from input data set) and whiteboard image number (from input data set sample) connected with '_'.

Deletion of whiteboard images in input data set is disabled to not interfere with real_world_data_collection.py until whiteboard data collection is finished.

### DIFF
--- a/marimbabot_vision/scripts/flatten_whiteboard_data.py
+++ b/marimbabot_vision/scripts/flatten_whiteboard_data.py
@@ -6,15 +6,6 @@ from shutil import copyfile
 
 import tqdm
 
-parser = argparse.ArgumentParser(
-                    prog = 'flatten_whiteboard_data.py',
-                    description = 'This script generates a separate dataset including one sample folder for each captured whiteboard picture. Each sample folder contains the correspoding .txt file, the .ly file and the .png file.')
-
-parser.add_argument('-o', '--outdir', type=str, required=True, help='The output directory for storing the whiteboard dataset')
-parser.add_argument('-i', '--indir', type=str, required=True, help='The input directory containing the captured whiteoard images')
-parser.add_argument('-p', '--prefix', type=str, default='staff_1', help='The prefix of the file names. Default: staff_1')
-parser.add_argument('-w', '--workers', type=int, default=30, help='The number of workers. Default: 30')
-
 def generate_sample_folder(sample):
 
     sample_folder = os.path.join(args.indir, sample)
@@ -27,7 +18,15 @@ def generate_sample_folder(sample):
         copyfile(os.path.join(sample_folder, real_im), os.path.join(real_sample_folder, f'{args.prefix}.png'))
 
 if __name__ == "__main__":
-    args = parser.parse_args()
+    parser = argparse.ArgumentParser(
+                        prog = 'flatten_whiteboard_data.py',
+                        description = 'This script generates a separate dataset including one sample folder for each captured whiteboard picture. Each sample folder contains the correspoding .txt file, the .ly file and the .png file.')
 
+    parser.add_argument('-o', '--outdir', type=str, required=True, help='The output directory for storing the whiteboard dataset')
+    parser.add_argument('-i', '--indir', type=str, required=True, help='The input directory containing the captured whiteoard images')
+    parser.add_argument('-p', '--prefix', type=str, default='staff_1', help='The prefix of the file names. Default: staff_1')
+    parser.add_argument('-w', '--workers', type=int, default=30, help='The number of workers. Default: 30')
+
+    args = parser.parse_args()
     with Pool(args.workers) as pool:
         list(tqdm.tqdm(pool.imap(generate_sample_folder, os.listdir(args.indir)), total=len(os.listdir(args.indir))))
--- a/marimbabot_vision/scripts/flatten_whiteboard_data.py
+++ b/marimbabot_vision/scripts/flatten_whiteboard_data.py
@@ -1,0 +1,25 @@
+import os
+import tqdm
+from multiprocessing import Pool
+import re
+from shutil import copyfile
+
+NUM_WORKER = 30
+INPUT_DIR = 'data'
+OUTPUT_DIR = 'data_wb'
+NAME_PREFIX = 'staff_1'
+
+def main(sample):
+    sample_folder = os.path.join(INPUT_DIR, sample)
+    for real_im in [i for i in os.listdir(sample_folder) if os.path.isfile(os.path.join(sample_folder,i)) and 'real' in i]:
+        num = re.search(f'{NAME_PREFIX}_real_(.*).png', real_im).group(1)
+        real_sample_folder = os.path.join(OUTPUT_DIR, f'{sample}_{num}')
+        os.makedirs(real_sample_folder, exist_ok=True)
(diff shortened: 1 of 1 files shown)


marimbabot_vision/scripts/flatten_whiteboard_data.py
```diff
@@ -0,0 +1,25 @@
...
+from multiprocessing import Pool
+import re
+from shutil import copyfile
+
+NUM_WORKER = 30
+INPUT_DIR = 'data'
+OUTPUT_DIR = 'data_wb'
+NAME_PREFIX = 'staff_1'
+
+def main(sample):
```
Juphex, line 12: I would prefer to rename this to a more descriptive name

marimbabot_vision/scripts/flatten_whiteboard_data.py
```diff
@@ -0,0 +1,25 @@
...
+    sample_folder = os.path.join(INPUT_DIR, sample)
+    for real_im in [i for i in os.listdir(sample_folder) if os.path.isfile(os.path.join(sample_folder,i)) and 'real' in i]:
+        num = re.search(f'{NAME_PREFIX}_real_(.*).png', real_im).group(1)
+        real_sample_folder = os.path.join(OUTPUT_DIR, f'{sample}_{num}')
+        os.makedirs(real_sample_folder, exist_ok=True)
+        copyfile(os.path.join(sample_folder, f'{NAME_PREFIX}.txt'), os.path.join(real_sample_folder, f'{NAME_PREFIX}.txt'))
+        copyfile(os.path.join(sample_folder, f'{NAME_PREFIX}.ly'), os.path.join(real_sample_folder, f'{NAME_PREFIX}.ly'))
+        copyfile(os.path.join(sample_folder, real_im), os.path.join(real_sample_folder, real_im))
+        # os.remove(os.path.join(sample_folder, real_im)) # !!! only after whiteboard data collection is finished !!!
```
Juphex, line 21: I would delete this line.
might be good to keep the "original" data anyways

Juphex:
maybe we want to add a file doc. Not only here but for any file because it seems that we have many small scripts we might loose track of later

marimbabot_vision/scripts/flatten_whiteboard_data.py
Juphex, line 12: I would recommend to add a comment on what this does and maybe why its needed

marimbabot_vision/scripts/flatten_whiteboard_data.py
```diff
@@ -0,0 +1,25 @@
+import os
```
Flova, line 1: Maybe sort the imports. There is an option for it in vs code if you right-click the code.

marimbabot_vision/scripts/flatten_whiteboard_data.py
```diff
@@ -0,0 +1,25 @@
...
+import tqdm
+from multiprocessing import Pool
+import re
+from shutil import copyfile
+
+NUM_WORKER = 30
```
Flova, line 7: The other scripts are now being migrated to argparse, maybe we want to do this here too.

marimbabot_vision/scripts/flatten_whiteboard_data.py
```diff
@@ -0,0 +1,25 @@
...
+INPUT_DIR = 'data'
+OUTPUT_DIR = 'data_wb'
+NAME_PREFIX = 'staff_1'
+
+def main(sample):
+    sample_folder = os.path.join(INPUT_DIR, sample)
+    for real_im in [i for i in os.listdir(sample_folder) if os.path.isfile(os.path.join(sample_folder,i)) and 'real' in i]:
```
Flova, line 14: This works. Another way would be the usage of `glob` which might be a bit more consice.

Flova (changes requested):

Flova:
I changed the output png naming in 0fda3279b5fa3946ef3d2d6f3aa188f748aafdcf to resemble the convention used else where.

Flova:
Ping @scaredycode

marimbabot_vision/scripts/flatten_whiteboard_data.py
scaredycode: Done. Thanks for the info!

marimbabot_vision/scripts/flatten_whiteboard_data.py
scaredycode: Yes, you are absolutely right. I added a description to argparse, hope that will do.

marimbabot_vision/scripts/flatten_whiteboard_data.py
scaredycode: Done. I agree, deleting the original files might be a really bad idea after all the work it took to capture each image.

marimbabot_vision/scripts/flatten_whiteboard_data.py
scaredycode: Done. Looks a lot fancier now :P

marimbabot_vision/scripts/flatten_whiteboard_data.py
scaredycode: You are right, I was a bit lazy here. I changed the name of this function to 'generate_sample_folder'.

marimbabot_vision/scripts/flatten_whiteboard_data.py
```diff
@@ -1,25 +1,33 @@
...
-        copyfile(os.path.join(sample_folder, real_im), os.path.join(real_sample_folder, f'{NAME_PREFIX}.png'))
-        # os.remove(os.path.join(sample_folder, real_im)) # !!! only after whiteboard data collection is finished !!!
+        copyfile(os.path.join(sample_folder, f'{args.prefix}.txt'), os.path.join(real_sample_folder, f'{args.prefix}.txt'))
+        copyfile(os.path.join(sample_folder, f'{args.prefix}.ly'), os.path.join(real_sample_folder, f'{args.prefix}.ly'))
+        copyfile(os.path.join(sample_folder, real_im), os.path.join(real_sample_folder, f'{args.prefix}.png'))
 
 if __name__ == "__main__":
-    with Pool(NUM_WORKER) as pool:
-        list(tqdm.tqdm(pool.imap(main, os.listdir(INPUT_DIR)), total=len(os.listdir(INPUT_DIR))))
+    args = parser.parse_args()
```
Flova, line 30: Did you check that everything works with this scoping? Because `args` is defined down here and not explicitly passed to the functions.

Also consider moving the parser definition into the if name == main thing.

Flova (approved):

Juphex (approved):
rebase or merge and we can forward this into main :)

marimbabot_vision/scripts/flatten_whiteboard_data.py
scaredycode: I tested it and it worked, but I was wondering about that too.
I put the parser definition on top as some sort of substitute for a script description, but I guess this is a kind of inelegant solution.

scaredycode: I moved the parser stuff to the main function now. Tested once more, still works.

PR #18, opened 2023-02-11
Fix tokenization

Flova:
Some fixes regarding whitespace issues in the network output.

### DIFF
--- a/marimbabot_vision/scripts/detect.py
+++ b/marimbabot_vision/scripts/detect.py
@@ -47,14 +47,7 @@ outputs = model.generate(
 )
 
 # Decode tokens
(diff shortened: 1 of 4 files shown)


Juphex:
whats fixed here exactly?

Flova:
- fixed some formatting
- use the proper way for the removal of special tokens (end of sequence, padding, ...)
- add all notes (e.g. `c''1`) to the tokenizer vocabulary. This solves some issues where the notes are split into subtokens. This results in unessesary many tokens (and autoregressive steps during the prediction), longer sequences during training and issues where the tokenizer inserts space tokens between subtokens of the note (e.g. `c''1` is `c `1`). I do not fully understand the reason for the last one but my fix works.

Juphex (approved):

PR #20, opened 2023-03-01
Feature/vision node

Juphex:
add camera node
add image processing


marimbabot_vision/launch/marimbabot.launch
```diff
@@ -3,5 +3,13 @@
      <rosparam>
        cv_camera/device_id: 0
      </rosparam>
+      <rosparam>
+       cv_camera/rate: 1
```
Flova, line 7: Idk but we could run the cam at a higher rate, so the delay is reduced a bit.

marimbabot_vision/launch/marimbabot.launch
```diff
@@ -3,5 +3,13 @@
...
+     </rosparam>
+</node>
+<node name="vision_node" pkg="marimbabot_vision" type="vision_node.py">
+     <rosparam>
```
Flova, line 11: I am not familiar with this notation for parameters. I would suggest a config folder in the package. We can place a yaml file with all parameters there and load it in the launch file

marimbabot_vision/src/vision_node.py
```diff
@@ -1,25 +1,85 @@
...
 
-def callbackImage(data):    # TODO
-    pass
-#  rospy.loginfo()
-    
-def listener():
+def detect_notes(img_array: np.array, pre_processor: DonutProcessor, model: VisionEncoderDecoderModel):
+    # Load image
+    image = PILImage.open(img_array).convert('RGB')
+
```
Flova, line 15: We might need to rescale the image at some point, because the camera image will be much more high res than the model input. It could be that the donut preprocessor is already doing that, but I would crosscheck this.

marimbabot_vision/src/vision_node.py
```diff
@@ -1,25 +1,85 @@
...
+        use_cache=True,
+        num_beams=1,
+        bad_words_ids=[[pre_processor.tokenizer.unk_token_id]],
+        return_dict_in_generate=True,
+    )
+
+    # Decode tokens
+    sequence = pre_processor.batch_decode(outputs.sequences)[0][4:]
+
+    # Clean output sequence
```
Flova, line 46: This is not nessesary anymore. The newest version of the detection script handles it via a flag in the decoding step.

marimbabot_vision/src/vision_node.py
```diff
@@ -1,25 +1,85 @@
...
+    rospy.Subscriber("cv_camera_node/image_raw", ROSImage, callbackImage, callback_args=(pre_processor, model))
 
     # spin() simply keeps python from exiting until this node is stopped
     rospy.spin()
 
 if __name__ == '__main__':
-    listener()
+    MODEL = './model'
+
+if rospy.has_param('MODEL_PATH'):
```
Flova, line 74: What is the else case?

Flova (changes requested):

marimbabot_vision/launch/marimbabot.launch
Juphex: 1 was just a safe bet. I can adjust that

marimbabot_vision/src/vision_node.py
Flova: Ah nvm. I looked at it on my phone, which messed with the indentation a bit.

marimbabot_vision/src/vision_node.py
Juphex: Yes, it should rescale it. From the docs I got that there is a parameter "do_rescale".

DonutProcessor is instance of DonutImageProcessor. There the default value for the parameter is set to True.

https://huggingface.co/docs/transformers/model_doc/donut#transformers.DonutImageProcessor

Flova: Okay nice 👍

marimbabot_vision/launch/marimbabot.launch
Flova: Okay

Flova: I would go for 5 in the beginning

marimbabot_vision/package.xml
```diff
@@ -14,8 +14,15 @@
   <buildtool_depend>catkin</buildtool_depend>
   <build_depend>rospy</build_depend>
   <build_depend>std_msgs</build_depend>
+  <build_depend>cv_camera</build_depend>
```
Flova, line 17: We also need to add the CV bridge dependency

marimbabot_vision/src/vision_node.py
```diff
@@ -0,0 +1,86 @@
...
+        use_cache=True,
+        num_beams=1,
+        bad_words_ids=[[pre_processor.tokenizer.unk_token_id]],
+        return_dict_in_generate=True,
+    )
+
+    # Decode tokens
+    sequence = pre_processor.batch_decode(outputs.sequences, skip_special_tokens=True)[0]
+
+    print(sequence)
```
Flova, line 46: These two lines print the output two times. I would remove them or at least lower the log level to debug. Otherwise the terminal gets spamed quite a bit and one can always easily subscribe to the output topic.

marimbabot_vision/src/vision_node.py
```diff
@@ -0,0 +1,86 @@
...
+def callbackImage(data: ROSImage):
+    rospy.logdebug("received img")
+
+    # http://docs.ros.org/en/api/sensor_msgs/html/msg/Image.html
+    bridge = CvBridge()
+
+    try:
+      cv_image = bridge.imgmsg_to_cv2(data, encoding="passthrough")
+    except CvBridgeError as e:
+      rospy.loginfo(e)
```
Flova, line 59: This would be the error loglevel

marimbabot_vision/src/vision_node.py
```diff
@@ -0,0 +1,86 @@
...
+
+    detect_notes(cv_image)
+
+    
+def listener(pre_processor, model):
+    rospy.init_node('vision_receiver', anonymous=True)
```
Flova, line 65: I would not launch this node as an anonymous one.

Flova (changes requested):

Juphex:
I have set up the project on a clean machine. Camera/Node works as expected - the parameters "rosparam" as well.
Did not test it with the model it self yet. However, should work as the code for that is from the script.

marimbabot_vision/launch/marimbabot.launch
Juphex: I don't know what you mean by that exactly. If you want you can create a pull request for that. :)

marimbabot_vision/src/vision_node.py
```diff
@@ -43,42 +43,43 @@ def detect_notes(open_cv_img, pre_processor: DonutProcessor, model: VisionEncode
...
-    print(sequence)
+    rospy.logdebug(sequence)
 
-    rospy.loginfo(sequence)
-
-def callbackImage(data: ROSImage):
+def callbackImage(data: ROSImage, callback_args):
     rospy.logdebug("received img")
 
+    pre_processor = callback_args[0]
```
Flova, line 51: You can rewrite this as `pre_processor, model = callback_args`

marimbabot_vision/launch/marimbabot.launch
```diff
@@ -1,10 +1,10 @@
 <launch>
 <node name="cv_camera_node" pkg="cv_camera" type="cv_camera_node" respawn="true">
      <rosparam>
-       cv_camera/device_id: 0
+       device_id: 0
```
Flova, line 4: This can be one rosparam block (or better config file) and we could add a flag if we want to launch without the camera (e.g. if we use an rosbag). Similarly we should make the camera device id configurable via an launch argument.

Flova (changes requested):

marimbabot_vision/launch/marimbabot.launch
Flova: Like this: https://github.com/bit-bots/bitbots_navigation/blob/848da868363ccd4201f6cbd0c86d1ce28bb8f69f/bitbots_localization/launch/localization.launch#L13
Which loads this parameter file:
https://github.com/bit-bots/bitbots_navigation/blob/ros1/bitbots_localization/config/config.yaml

Flova:
> I have set up the project on a clean machine. Camera/Node works as expected - the parameters "rosparam" as well. Did not test it with the model it self yet. However, should work as the code for that is from the script.

Nice!

marimbabot_vision/launch/marimbabot.launch
Flova: This is especially relevant for laptop with the USB webcam and an internal one. Editing the launch file for this is not so nice.

marimbabot_vision/launch/marimbabot.launch
Juphex: In the case I don't do it in near future. Would you mind merging this first and adjusting that later? I could create a new PR for that.

Flova (approved):

PR #23, opened 2023-03-07
add separate config

Juphex:
I guess that is faster than you expected 😆  @Flova


Flova (approved):

Flova:
We could still add an launch argument for the camera device.

PR #24, opened 2023-03-08
add lilypond visualization

Juphex:
opens a visualization of the notes recognized by the vision node

### DIFF
--- a/marimbabot_vision/src/vision_node.py
+++ b/marimbabot_vision/src/vision_node.py
@@ -65,9 +65,10 @@ def callbackImage(data: ROSImage, callback_args):
     sentence_publisher.publish(sentence)
 
     
-def listener(pre_processor, model, sentence_publisher):
+def listener(pre_processor, model):
     rospy.init_node('vision_receiver')
 
+    sentence_publisher = rospy.Publisher("recognized_sentence", String)
     rospy.Subscriber("cv_camera_node/image_raw", ROSImage, callbackImage, callback_args=(pre_processor, model, sentence_publisher))
 
     # spin() simply keeps python from exiting until this node is stopped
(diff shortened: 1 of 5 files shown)


marimbabot_vision/src/vision_node.py
```diff
@@ -58,13 +61,14 @@ def callbackImage(data: ROSImage, callback_args):
     except CvBridgeError as e:
       rospy.logerror(e)
 
-    detect_notes(cv_image, pre_processor, model)
+    sentence = detect_notes(cv_image, pre_processor, model)
+    sentence_publisher.publish(sentence)
 
     
-def listener(pre_processor, model):
+def listener(pre_processor, model, sentence_publisher):
```
Flova, line 68: What is the purpose of this function? I think the split is not necessary, and you can integrate it into the other part below.

marimbabot_vision/src/vision_node.py
```diff
@@ -58,13 +61,14 @@ def callbackImage(data: ROSImage, callback_args):
     except CvBridgeError as e:
       rospy.logerror(e)
 
-    detect_notes(cv_image, pre_processor, model)
+    sentence = detect_notes(cv_image, pre_processor, model)
```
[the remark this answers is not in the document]
Juphex: That is exactly what I have implemented

Juphex:
I need to check whether the picture is shown and then it can be reviewed

marimbabot_vision/src/visualization_node.py
```diff
@@ -0,0 +1,51 @@
...
+import numpy as np
+import rospy
+from abjad import Block, LilyPondFile
+from abjad.persist import as_png
+from cv_bridge import CvBridge
+from sensor_msgs.msg import Image as ROSImage
+from std_msgs.msg import String
+
+
+def callbackVisionResults(sentence, args):
```
Flova, line 15: We normally use snake_case, but it is fine imo.

Flova (approved):
LGTM, if you tested it and it works.

Juphex:
I tested it in Rviz, what is what I got. 
Rviz: 
imageimage

What you sent in our chat @Flova : 
image

PR #26, opened 2023-03-09
Sequence motion planning

Flova:


marimbabot_planning/src/move_to_pose.cpp
```diff
@@ -278,6 +282,7 @@ int main(int argc, char **argv)
     // Hit a sequence of points in cartesian space left and right of "pose"
 
     // Define hit points
+<<<<<<< Updated upstream
```
Flova, line 285: It seems like you pushed an unresolved merge

Flova: This is also the reason for the CI failure.

marimbabot_planning/src/move_to_pose.cpp
```diff
@@ -24,15 +24,16 @@ moveit::planning_interface::MoveGroupInterface::Plan concatinated_plan(std::vect
     assert(plans.size() > 0);
 
     moveit::planning_interface::MoveGroupInterface::Plan plan{plans[0]};
-    ros::Duration time_from_start{0};
+    ros::Duration time_from_start{1};
```
Flova, line 27: Why did you do this? @berkgungor semantically zero is correct. Also, what is `time_added`?

berkgungor (approved):
LGTM!!!

PR #33, opened 2023-03-30
Feature/marimba ur5 integration

berkgungor:


marimba_ur5_moveit_config/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+cmake_minimum_required(VERSION 3.1.3)
+project(marimbabot_ur5_moveit_config)
```
Juphex, line 2: rename the folder from "marimba_ur5_moveit_config" to "marimbabot_ur5_moveit_config" for consistency ?

Juphex: also I guess that it is best practice that project and folder name match

Juphex (changes requested):
do you use all files/are all files necessary?

like warehouse.launch, demo.launch

marimbabot_description/package.xml
```diff
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 <package format="2">
-  <name>marimbabot_sim</name>
+  <name>marimbabot_description</name>
   <version>0.0.0</version>
-  <description>The marimbabot_sim package</description>
+  <description>The marimbabot_description package</description>
```
Flova, line 5: Add more detailed description

Juphex (commented):
except that I disagree with the boilerplate moveit gives, the rest
lgtm when the requested changes by the others are applied

marimbabot_ur5_moveit_config/.setup_assistant
```diff
@@ -0,0 +1,11 @@
+moveit_setup_assistant_config:
+  URDF:
+    package: marimbabot_description
+    relative_path: urdf/marimbabot_ur5.urdf.xacro
+    xacro_args: ""
```
berkgungor, line 5: we can remove xacro_args since it's empty stirng

Flova: I think it is a nessesary argument even if it is empty

Flova (approved):

PR #34, opened 2023-04-05
Adds Action Server for connecting behaviour and planning

### LINKED ISSUE #50: connect detected notes with move_to_pose

Juphex:
Closes #50


ateRstones:
I would maybe think about using a header file or a class definition stub as the layout has become quite confusing especially with the private block at the end.
Otherwise this looks good to me, but I don't have deep knowledge here

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,110 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
...
+    ros::Subscriber sub_;
+
+};
+
+int main(int argc, char **argv)
+{
+    ros::init(argc, argv, "marimba_move_lilypond_input_viz");
+
+    // Create SubscribeAndPublish object
+    SubscribeAndPublish SAPObject;
```
ateRstones, line 211: I would write this object name in snake case to be consistent with all the other code. Also at least from my understanding please don't write variables starting with upper case

Juphex:
Also closes https://github.com/UHHRobotics22-23/marimbabot/issues/50

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,110 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
...
-
-        // Add lifetime (duration of the note)
-        marker.lifetime = ros::Duration(std::get<2>(hits[i]));
-
-        // Publish the marker
-        hit_marker_pub.publish(marker);
-
-        // Check if this is not the last one
-        if(i < hits.size() - 1)
+        ROS_INFO("Converted lilypond sequence to cartesian poses and times");
```
Flova, line 144: Debug log level

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,110 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
...
-        );
+        ROS_INFO("Starting marimba_move");
 
-    ROS_INFO("Converted lilypond sequence to cartesian poses and times");
+        std::string lilypond_string = lilypond_sentence->data.c_str();
 
-    // Loop over all hits and publish a marker at the correct time
-    double time = 0.0;
+        // Convert lilypond sequence to cartesian poses and times
+        auto hits = lilypond_to_cartesian(
```
Flova, line 138: Optimally the result of this function would be directly published and the viz is a different node.

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,110 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
 }
 
 
-int main(int argc, char **argv)
+
+class SubscribeAndPublish
```
Flova, line 111: Consider changing the class name to something like `LilypondMotionConverter`

Flova (changes requested):

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,110 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
...
-    spinner.start();
+public: 
+    SubscribeAndPublish() {
+        //Topic you want to publish
+        hit_marker_pub = n_.advertise<visualization_msgs::Marker>("hit_marker", 1);
 
-    // Create publisher for the hit marker
-    ros::Publisher hit_marker_pub = node_handle.advertise<visualization_msgs::Marker>("hit_marker", 1);
+        //Topic you want to subscribe
+        sub_ = n_.subscribe("vision_node/recognized_sentence", 1, &SubscribeAndPublish::moveToPoseCallback, this);
```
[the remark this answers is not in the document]
Juphex: yes, see https://github.com/UHHRobotics22-23/marimbabot/blob/cf00df26cc284b9e001c1f992fccc244c5fa889d/marimbabot_vision/src/vision_node.py#L67

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,110 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
...
 
-    ROS_INFO("Converted lilypond sequence to cartesian poses and times");
+        std::string lilypond_string = lilypond_sentence->data.c_str();
 
-    // Loop over all hits and publish a marker at the correct time
-    double time = 0.0;
+        // Convert lilypond sequence to cartesian poses and times
+        auto hits = lilypond_to_cartesian(
+            tfBuffer,
+            "base_link",
```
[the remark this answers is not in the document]
Juphex: you are right

marimbabot_planning/src/lilypond_to_hitpoint.cpp
Juphex: https://github.com/UHHRobotics22-23/marimbabot/issues/63

Juphex:
Closes https://github.com/UHHRobotics22-23/marimbabot/issues/58 as well

marimbabot_planning/src/lilypond_to_hitpoint.cpp
```diff
@@ -106,87 +107,111 @@ std::vector<std::tuple<geometry_msgs::PoseStamped, double, double>> lilypond_to_
...
-int main(int argc, char **argv)
+/**
+ * @brief ROS node that subscribes to the recognized lilypond sequence and publishes a marker at the correct time
+ * class is needed to access the publisher inside the callback function
+*/
+class LilypondMotionConverter
 {
-    ros::init(argc, argv, "marimba_move_lilypond_input_viz");
-    ros::NodeHandle node_handle;
-    ros::AsyncSpinner spinner(1);
```
berkgungor, line 113: would be nice to keep AsyncSpinner for threads

Juphex: done

marimbabot_planning/src/actions/hit_sequence.action
```diff
@@ -0,0 +1,8 @@
+#goal definition
+int32[] sequence
```
Flova, line 2: We should add a tempo to the goal

Flova: E.g. one int with bmp

Flova (changes requested):
I would move the message/action/service things to a package called `marimbabot_interfaces` or `marimbabot_msgs`.

marimbabot_planning/src/actions/hit_sequence.action
```diff
@@ -0,0 +1,8 @@
+#goal definition
```
v4hn, line 1: What is this `sequence` supposed to represent? It seems an insufficient representation either way if you want to support playing multiple notes with different durations in a score.

Assuming each int is basically a midi tone, you would only have to add an onset time per note to get a generic "midi-style" representation.

Additionally, you need to add a comprehensible description because this is the only text that defines what exactly the fields mean. (The implementation has to implement what the definition says, not the other way around :))

marimbabot_planning/CMakeLists.txt
```diff
@@ -29,3 +32,13 @@ target_link_libraries(lilypond_to_hitpoint ${catkin_LIBRARIES})
 install(TARGETS planning lilypond_to_hitpoint
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
+
+# Action 
+add_action_files(
+  DIRECTORY src/actions
+  FILES hit_sequence.action
```
v4hn, line 39: [By convention](http://wiki.ros.org/ROS/Patterns/Conventions) message/service/action types should be `CamelCase`

marimbabot_planning/src/actions/hit_sequence.action
```diff
@@ -0,0 +1,8 @@
...
+int32[] sequence
+---
+#result definition
+int32[] sequence
+---
+#feedback
+bool worked
```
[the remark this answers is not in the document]
v4hn: Feedback is something you send continuously while the action executes. What would `worked = false` and `worked = true` mean when you send it more than once during playing?

marimbabot_planning/src/actions/hit_sequence.action
```diff
@@ -0,0 +1,8 @@
+#goal definition
+int32[] sequence
+---
+#result definition
+int32[] sequence
```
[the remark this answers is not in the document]
v4hn: actual time of hit note + a success field would also make sense.

marimbabot_planning/src/actions/hit_sequence.action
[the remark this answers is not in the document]
Flova: Oh nvm I thougth this was the result part. The ideas was to have the "number of played notes" as a continous feedback and the `success` as the result.

marimbabot_planning/src/actions/hit_sequence.action
Flova: For the representation discussion see #73. Imo. it could either be the lilypond string or a sequence of hit points with timings. MIDI onsets would also be an option, but it is not really used in our code currently. Either way, I think ints are not a suitable representation here.

v4hn: > MIDI onsets would also be an option, but it is not really used in our code currently

Maybe I formulated that wrongly when I mentioned it yesterday: You definitely don't want to use MIDI as interface format (although we even have [some level of support for that](https://github.com/TAMS-Group/tams_midi_publisher) from prior work). But if you send 1. timing 2. a [midi-note number](http://midi.teragonaudio.com/tech/midispec/noteon.htm) (corresponding to western musical notation half-tones around mid-C with value 60) and a "velocity" or "loudness" parameter for each hit, that is essentially the information stored in a Note-Ons of a midi-file as well :-)

Juphex: I implemented an int for bpm and list of strings for the notes

marimbabot_planning/src/actions/hit_sequence.action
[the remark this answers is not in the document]
Juphex: I removed the parameter and added "success" to the result definition

marimbabot_planning/src/actions/hit_sequence.action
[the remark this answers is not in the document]
Juphex: not quite sure about the time of hit note but I added it as a string and am ready for complains

marimbabot_planning/src/actions/hit_sequence.action
Juphex: You are right, I changed it to a list of strings for now.

marimbabot_planning/CMakeLists.txt
Juphex: done

Juphex:
As it is now, the ActionServer receives a LilyPond String as an Action Goal and forwards it to "lilypond_to_hitpoint.cpp".
@berkgungor @Flova  @scaredycode 
We should get together and 
1. talk about client side (ActionClient), that should be straight forward
2. talk about the robot side, i.e. the data representation that is needed/suitable in order to execute a music sequence including dynamics, note values and bpm maybe. This assumes that Lilypond Parsing (to a MIDI representation or whatever) happens inside the ActionServer or somewhere later. :)

Juphex:
TODO adjust `lilypond_to_hitpoint.cpp`  in order to accept lilypond dynamics

marimbabot_msgs/actions/HitSequence.action
```diff
@@ -0,0 +1,15 @@
+#goal definition
+std_msgs/String[] notes_sequence # list of note string, e.g. c4
```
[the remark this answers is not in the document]
Juphex: We decided on sending a lilypond string, thus, everything that is considered in our data set (including tempo, slurs etc.) should be possible to extract from there. Considering the current feedback and result (played_sequence, number of played notes), as it is defined to date, what we have in https://github.com/UHHRobotics22-23/marimbabot/blob/main/marimbabot_planning/src/lilypond_to_hitpoint.cpp comes in handy because it already uses a a list for each hit/bar on the marimba. What are your opinions on that?

v4hn:
> What are your opinions on that?

You can pass the goal as a string of course, as we discussed (at length) before.
You have to be exact about what strings you accept though because strings can contain anything (e.g., json wrapped base64 encoding of cat pictures...) and thus do not define an interface.
So if you do not want to support *all* of lilypond you should document examples you want to support above.

To illustrate various examples of features that I think are easy to support through lilypond or that you need to support to some level due to requirements, I wrote a (not-so-)[small example using the first phrase of "all my ducklings"](https://github.com/TAMS-Group/music_perception/blob/main/ly).
A lot of the `.ly` file is for formatting and does not affect the sound, so the actual features are not that intimidating :rabbit2: 

As I mentioned to @Flova and @scaredycode last week, I also wrote some code for myself [to convert a lilypond string/file into a list of "audio-perception-style" NoteOnsets](https://github.com/TAMS-Group/music_perception/blob/main/scripts/ly2piece.py). That is basically what you would expect the music perception to output when you play the piece. It calls lilypond in the background and thus would support all of lilypond-syntax, but I don't know whether you want to use it at all or possibly restrict the syntax on purpose to make it easier to interpret.

Juphex:
I defined the action and a new message type.
Naming and definition strict to what you have written down there @scaredycode  https://github.com/UHHRobotics22-23/marimbabot/issues/111 . 
The overall structure should be ready now just the TODOs, like creating a publisher and subscriber for information exchange between planning and action server, have to be implemented.

marimbabot_planning/include/marimbabot_planning/planning.h
```diff
@@ -1,92 +1,91 @@
...
+#include <geometry_msgs/Pose.h>
 #include <iostream>
+#include <marimbabot_msgs/HitSequenceAction.h>
+#include <marimbabot_planning/utils.h>
 #include <moveit_msgs/DisplayTrajectory.h>
 #include <moveit_msgs/RobotState.h>
 #include <moveit_msgs/RobotTrajectory.h>
 #include <moveit/move_group_interface/move_group_interface.h>
 #include <moveit/robot_state/conversions.h>
+#include <regex>
```
Juphex, line 13: Do we use this? I did not find it 😅

marimbabot_planning/include/marimbabot_planning/planning.h
```diff
@@ -1,92 +1,91 @@
...
 #include <ros/ros.h>
 #include <tf2_geometry_msgs/tf2_geometry_msgs.h>
 #include <tf2_ros/transform_listener.h>
 #include <tf2/LinearMath/Quaternion.h>
 #include <trajectory_msgs/JointTrajectoryPoint.h>
```
Juphex, line 18: this moved to utils, right?

PR #36, opened 2023-04-07
Add optimization notebook

Flova:
Add a notebook that shows how we could optimize the marimba's pose based on audio feedback.

The notebook only covers the optimization side of things and leaves out all the actual motion, model and audio interfacing.

We can build a self calibration node from that.


marimbabot_planning/scripts/self_calibration_optimization_example.ipynb
```diff
@@ -0,0 +1,355 @@
...
+    "                  [0, 0, 1]])\n",
+    "\n",
+    "# Transforms the bars of the marimba model according to a given transformation matrix in a vectorized way\n",
+    "def transform_model(lines, T):\n",
+    "    # Extract the rotation and translation components from the transformation matrix\n",
+    "    R = T[:2, :2]\n",
+    "    t = np.hstack((T[:2, 2], [0]))\n",
+    "    \n",
+    "    # Apply the rotation to the model\n",
+    "    rotated_model = np.einsum('ijk,lk->ijl', lines, T)\n",
```
Juphex, line 123: what does the argument mean?

Juphex:
So, you want to optimize the sound / hit of a bar / note using its audio information?

marimbabot_planning/scripts/self_calibration_optimization_example.ipynb
Flova: https://en.wikipedia.org/wiki/Einstein_notation

Flova:
> So, you want to optimize the sound / hit of a bar / note using its audio information?

Correct. This only includes a demo of the optimization using synthetic data, but we can build a full calibration node from it.

marimbabot_planning/scripts/self_calibration_optimization_example.ipynb
Flova: It is a short notation for what is essentially a batched matrix multiplication.

marimbabot_planning/scripts/self_calibration_optimization_example.ipynb
```diff
@@ -0,0 +1,355 @@
...
+   "outputs": [],
+   "source": [
+    "# Create evenly spaced bars on the marimba (should be taken from urdf in the future)\n",
+    "def generate_marimba_model(num_bars, key_length, key_width):\n",
+    "    # Create evenly spaced keys using NumPy's linspace function\n",
+    "    keys = np.linspace(0, key_width * num_bars, num=num_bars)\n",
+    "\n",
+    "    # Create lines based on the key points \n",
+    "    lines = np.zeros((num_bars, 2, 3))\n",
+    "    lines[:, :, 0] = keys[:, None]\n",
```
ateRstones, line 50: I was confused by the None in this line.
After some googling I understand it adds a new dimension, but maybe I would suggest using numpy.newaxis here as it makes the operation more readable. https://numpy.org/doc/stable/user/basics.indexing.html#dimensional-indexing-tools
Maybe this is nit picking though

ateRstones:
A further comment, I'm not sure if I find the location of the notebook appropriate as it is within a ros module. For me these should contain only the ros related code and not something like this. Or maybe in a different folder than where you would expect to find the ros components.

marimbabot_planning/scripts/self_calibration_optimization_example.ipynb
Flova: I think it is preference, but I like the broadcasting notation with `None`, but I can see your point.

Flova:
> A further comment, I'm not sure if I find the location of the notebook appropriate as it is within a ros module. For me these should contain only the ros related code and not something like this. Or maybe in a different folder than where you would expect to find the ros components.

I placed it in the `scripts` folder which can imo. contain non ROS scripts in contrast to the `src` folder. This is also the case in `marimbabot_vision`, where all the non ROS training and dataset code is located there.

Juphex (approved):

PR #46, opened 2023-04-13
change name of the repository from robot_project to marimbabot

Juphex:


ateRstones:
Sounds like a good idea, github should provide redirects for old urls too

Flova (approved):

Juphex:
That is exactly why I did this

PR #54, opened 2023-04-13
add the planning launch files to the bringup launch file

Juphex:


marimbabot_planning/launch/marimbabot.launch
```diff
@@ -0,0 +1,4 @@
+<launch>
+    <node name="lilypond_to_hitpoint" pkg="marimbabot_planning" type="lilypond_to_hitpoint.cpp"/>
```
Flova, line 2: I don't know if this should be in its own node in the end or if it should be included as a library in the motion.

marimbabot_planning/launch/marimbabot.launch
```diff
@@ -0,0 +1,4 @@
...
+    <node name="move_to_pose" pkg="marimbabot_planning" type="move_to_pose.cpp"/>
```
Flova, line 3: Did you confirm that this works, because I don't think that the `.cpp` is right from compiled cpp binaries in the ROS context.

Flova (changes requested):

marimbabot_planning/launch/marimbabot.launch
Juphex: sure, you are right

marimbabot_planning/launch/marimbabot.launch
berkgungor, line 2: node name defined as "move_lilypond_to_hitpoints" in the lilypond_to_hitpoint.cpp. Would be nice to keep the node names as same.

marimbabot_planning/launch/marimbabot.launch
berkgungor, line 3: Same for here. node name defined as "marimba_move" in the move_to_pose.cpp. Would be nice to keep the node names as same.

berkgungor (dismissed):

marimbabot_planning/launch/marimbabot.launch
Flova: I don't think this issue has been addressed.

marimbabot_planning/launch/marimbabot.launch
Flova: I think lilypond_to_hitpoint.cpp isn't correct ether

berkgungor: I debated too but here they included .py extension for the node.
http://wiki.ros.org/roslaunch/XML

Flova: This is the case because the py file is executed as is. The cpp file gets compiled and the resulting binary is named `lilypond_to_hitpoint` most of the time.

marimbabot_planning/launch/marimbabot.launch
Juphex: now it should be @Flova

Flova (approved):

PR #17, opened 2023-02-02
Feature/audio

yunlong-wang-cn:
currently, I am work on adding the intensity and duration of the tone, so please don't merge, until those have be done.

PR #74, opened 2023-04-20
add speech synthesis node

scaredycode:
adds speech synthesis node that listens to the response text topic and reads each message out loud using mimic3 and aplay


Juphex (commented):
you may cut out the commented lines inside the CMakesLists.txt :)
other than that looks fine for me

ateRstones:
Looks good to me

v4hn (commented):
mixing this node in with `_audio` seems questionable, because [the package description indicates this is meant for *music*/note perception](https://github.com/UHHRobotics22-23/marimbabot/blob/ec87b15253a60ac091cda245e137053cdee68493/marimbabot_audio/package.xml#L6-L8). To this end it creates unnecessary conflicts with #17.

Another thing I want to point out (I already mentioned it to @scaredycode) is that you never see the actual audio data sent to ALSA inside ROS, so you can't have another node, for example, attempt to filter the replayed audio from the microphone recording sound for note detection and you cannot easily record the audio output together with everything else [in a rosbag](http://wiki.ros.org/rosbag/). There's no reason to change anything in that regard, just be aware of the limitations.

Flova (commented):
I would move this to `marimbabot_speech` or something like this.

requirements.txt
```diff
@@ -7,4 +7,5 @@ Pillow==9.4.0
 pytorch_lightning==1.8.1
 torch==1.13.1
 tqdm==4.64.0
-transformers[sentencepiece]==4.26.0
\ No newline at end of file
+transformers[sentencepiece]==4.26.0
+mycroft-mimic3-tts==0.2.4
```
Juphex, line 11: sort please :)

marimbabot_speech/src/speech_synthesis.py
```diff
@@ -0,0 +1,39 @@
...
+        self.speed = "1.0"
+
+        # listens to the defined text responses from behavior node
+        self.response_sub = rospy.Subscriber("behavior_node/response", String, self.callback_response)
+
+    def callback_response(self, response):
+        text = response.data
+
+        mimic_subprocess = subprocess.Popen(
+            ('mimic3', '--voice', self.voice, '--length-scale', self.speed, text),
```
Juphex, line 21: didn't you want to change this?

yunlong-wang-cn (commented):
From my side, overall, it is pretty good, but I don't know why you delete the CMake file from package marimbabot_audio

Juphex:
@yunlong-wang-cn I don't know who deleted it but it makes sense because we have no src / nodes inside marimbabot_audio within this pull request

yunlong-wang-cn:
@Juphex   I think we have that folder in the main branch before we merge this pull&request. Because in this pull&request the folder is deleted and this deletion is also merged into the main branch.  But it doesn't matter, we can just keep it. It doesn't influence anything.  hahaha, I am just wondering.

Flova:
I think marimba bot audio got renamed instead of copied when marimbabot speech was created.

yunlong-wang-cn:
Just a friendly reminder, please don't touch the part not belong to you, otherwise there will be a guy have a new conflict to resolve. :)

Issue #77, opened 2023-04-25
Create URDF for compliant mallet holder


cvhex:
We need a URDF model of the Mallet Holder to run simulations and test the robot. We have decided not to include compliance in the model, as this is very difficult to predict and could contain its own research area. Therefore the URDF model is kept very simple.

cvhex:
I have created a URDF model that can be intigrated. However, at the moment I still lack the Colisions. I will add them until thursday.

PR #86, opened 2023-04-27
URDF model for a flexibel mallet holder

cvhex:
#77  An updated URDF model for the flexible mallet holder. The design fits the Thomann VMR2.


ateRstones (approved):
LGTM (change top to head though)

PR #112, opened 2023-05-13
Adjusted hardware limits as the servo is only able to cover 180°

### LINKED ISSUE #107: Fix range of servo hardware adapter
When creating the urdf for the 2 mallet holder I noticed that the range used for the servo control is incorrectly converted from and to radians (factor of 2)
This has to be adjusted
https://github.com/UHHRobotics22-23/marimbabot/pull/112

ateRstones:
Closes #107

### DIFF
--- a/marimbabot_hardware/src/servo_interface.cpp
+++ b/marimbabot_hardware/src/servo_interface.cpp
@@ -98,7 +98,7 @@ bool ServoInterface::try_open_serial_port() {
             this->top_limit = atoi(top_limit_str.c_str());
             this->bottom_limit = atoi(bottom_limit_str.c_str());
             this->resolution = atoi(resolution_str.c_str());
-            this->radian_limit = (2 * M_PI) * ((this->top_limit - this->bottom_limit) / ((double) this->resolution));
+            this->radian_limit = M_PI * ((this->top_limit - this->bottom_limit) / ((double) this->resolution));
 
             ROS_INFO_STREAM("Servo controller: Loaded limits from arduino");
 
--- a/marimbabot_hardware/urdf/two_mallet_holder.urdf
+++ b/marimbabot_hardware/urdf/two_mallet_holder.urdf
@@ -23,6 +23,6 @@
         <origin xyz="0.0 0.0 0.0292"/>
         <parent link="mallet_neck"/>
         <child link="mallet_top"/>
-        <limit lower="0.0" upper="1.602" effort="1.0" velocity="7.0"/>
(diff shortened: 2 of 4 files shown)


marimbabot_hardware/src/servo_interface.cpp
```diff
@@ -98,7 +98,7 @@ bool ServoInterface::try_open_serial_port() {
             this->top_limit = atoi(top_limit_str.c_str());
             this->bottom_limit = atoi(bottom_limit_str.c_str());
             this->resolution = atoi(resolution_str.c_str());
-            this->radian_limit = (2 * M_PI) * ((this->top_limit - this->bottom_limit) / ((double) this->resolution));
+            this->radian_limit = M_PI * ((this->top_limit - this->bottom_limit) / ((double) this->resolution));
```
Flova, line 101: I am not sure about this formula. why is there pi in there? What is the unit of the limits?

berkgungor: not related to this modification but, catkin throws error due to arduino's serial library in servo_interface.cpp and somehow it is merged to main branch. Did you also modify cmake file after you add serial library ?

ateRstones: @berkgungor are you sure you reran rosdep? serial is specified in the package.xml but you have to install it with rosdep as it is a new dependency

marimbabot_hardware/urdf/two_mallet_holder.urdf
```diff
@@ -23,6 +23,6 @@
         <origin xyz="0.0 0.0 0.0292"/>
         <parent link="mallet_neck"/>
         <child link="mallet_top"/>
-        <limit lower="0.0" upper="1.602" effort="1.0" velocity="7.0"/>
+        <limit lower="0.0" upper="0.801" effort="1.0" velocity="7.0"/>
```
Juphex, line 26: This "upper" value is a value that you came up with, right?

Juphex: e.g. by experimenting

marimbabot_hardware/src/servo_interface.cpp
Juphex: @berkgungor Is it solved? I guess it was just merged locally

berkgungor: serial library requires manual installation on locals, catkin build throws error after rosdep update. Problem is solved after manual installation but would be nice if noted on read.me or somewhere.

berkgungor (approved):

Issue #121, opened 2023-05-25
Update marimba URDF for self coliisions


iibrahimli:
Update the `marimba.urdf.xacro` to avoid any objects intersecting, and set gazebo selfCollision to `true` for everything in `marimba.urdf.xacro`. This way, the marimba and ur5 can be loaded as one URDF and gazebo will be aware of collisions between the arm/mallet and the marimba.

PR #124, opened 2023-06-01
#123 Load marimba and arm setup separately in Gazebo

### LINKED ISSUE #123: Load marimba and arm setup separately in Gazebo
Allows to check for collisions between the arm + mallet and marimba. When they are loaded as part of the same URDF (as we did before), they are parts of the same object and their collisions are ignored if `selfCollide` field is `false`. An alternative solution would be #121.

iibrahimli:
Allows to check for collisions between the arm + mallet and marimba. When they are loaded as part of the same URDF (as we did before), they are parts of the same object and their collisions are ignored if `selfCollide` field is `false`. An alternative solution would be #121.
Closes #123


iibrahimli:
#127 implemented instead of this. Closing

PR #127, opened 2023-06-01
Enable self-collisions in Gazebo

### LINKED ISSUE #121: Update marimba URDF for self coliisions

iibrahimli:
Enable self-collisions in Marimba UR5 URDF

Also (unrelated change, I know) reverts `load_robot_description.launch` to `marimbabot_ur5_upload.launch` to keep it consistent with TAMS naming.

Closes #121


marimbabot_description/urdf/marimba.urdf.xacro
```diff
@@ -130,12 +130,12 @@ Changelog:
...
     </link>
 
-    <xacro:if value="$(arg bars_fixed_joint)">
+    <xacro:if value="${arg bars_fixed_joint}">
         <xacro:tams_fixedjoint name="bar_${note}/joint" parent="${parent}" child="bar_${note}" >
             <origin xyz="${pos}" rpy="0 0 0" />
         </xacro:tams_fixedjoint>
     </xacro:if>
-    <xacro:unless value="$(arg bars_fixed_joint)">
+    <xacro:unless value="${arg bars_fixed_joint}">
```
Juphex, line 138: Please revert. Doesn't work with curly brackets. At least not for me 😅

marimbabot_description/urdf/marimba.urdf.xacro
```diff
@@ -130,12 +130,12 @@ Changelog:
     </collision>
     </link>
 
-    <xacro:if value="$(arg bars_fixed_joint)">
+    <xacro:if value="${arg bars_fixed_joint}">
```
Juphex, line 133: same

marimbabot_description/urdf/marimba.urdf.xacro
Juphex: At least when I start it in rviz.

Juphex (changes requested):
As I understood the complete model is now enabled to have a collision. Doesn't it make more sence to enable it only for specific parts like the bars? at least for the start

iibrahimli:
> As I understood right now: the complete model is now enabled to have a collision. Doesn't it make more sence to enable it only for specific parts like the bars? at least for the start

@Juphex In that case mallet/arm would go right through side boxes or body, which is not the desired behavior.

Juphex (approved):

Juphex:
works

PR #133, opened 2023-06-07
Feature/motion speed new

berkgungor:
- hitting speed modified with last motion updates 
- will be tested


marimbabot_planning/src/planning.cpp
```diff
@@ -236,9 +263,10 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
             move_group_interface_.getPlanningFrame(),
             tf_buffer_
         );
+        std::array<std::tuple<std::string, int32_t, ros::Time, ros::Duration, double>, 5> hit_speeds;
```
Flova, line 266: The `CartesianHitSequenceElement` contains all of this information. See [this](https://github.com/UHHRobotics22-23/marimbabot/blob/fdeb70c49f8464fac24aaefaee79ce53b5e1310a/marimbabot_msgs/msg/HitSequenceElement.msg) which is a part of said data structure as defined [here](https://github.com/UHHRobotics22-23/marimbabot/blob/fdeb70c49f8464fac24aaefaee79ce53b5e1310a/marimbabot_planning/include/marimbabot_planning/utils.h#L31).

berkgungor:
Superseeded by #208

PR #129, opened 2023-06-01
Bar hit detection in simulation

iibrahimli:
Detect bar contact based on joint velocity
Adds a node `detect_bar_contact` and a launchfile to run RViz, Gazebo, detect_bar_contact, and plotjuggler:
```
roslaunch marimbabot_simulation marimbabot_contact_debug.launch
```
The node `detect_bar_contact` detects which bars (notes) were hit based on `joint_states` and publishes a note string (e.g. `b5`, `dis5`) to topic `sim_notes`. The node `play_notes` then plays the detected notes using a crude (for now) approach with real WAV files recorded in a very noisy room.

A MIDI file is generated and saved to `marimbabot_simulation/midi` when the nodes are stopped.

Closes #163
Closes #41
Closes #104

PR #135, opened 2023-06-08
Separate simulation package

### LINKED ISSUE #134: Separate simulation into a package
Currently, Gazebo simulation files are in the `marimbabot_ur5_moveit_config` package. It would be clearer if they were in a separate package, e.g. `marimbabot_simulation`

iibrahimli:
Having a separate simulation package makes more sense at this point, instead of having a Gazebo launchfile in moveit config package. The sim package will contain:
- gazebo launchfile
- node(s) for detecting bars in sim & making appropriate sound (will be added in #129)

Closes #134


berkgungor (approved):
LGTM

PR #142, opened 2023-06-08
Feature/command recognition

yunlong-wang-cn:
Create a new PR to:

1. talk about merge the command_recognition_node and speech_node together, as speech_node. Since tts(text to speech) is a simple feature and in command_recognition_node a lot of stuff about speech recognition has been done. 
2. I find a new package gtts(google text to speech), it sounds better(less steel) than the one we currently use in the speech node. I plan to use it, the only con is it will generate a temp file, Maybe not elegant, but perhaps fine.
3. then i will change the repo according to what we talk and try to merge it.


marimbabot_command_recognition/src/text2speech_client.py
```diff
@@ -0,0 +1,13 @@
+from marimbabot_command_recognition.srv import Speak, SpeakResponse
+import rospy
+
+# only for show how to call the service of text2speech service
```
Juphex, line 4: could you move it into an examples or scripts folder, so it is clear what script is for testing/demonstrations

marimbabot_command_recognition/src/__init__.py
Juphex, line 1: since everything in /src/ is a node this isn't neccessary, isn't it?

requirements.txt
```diff
@@ -16,4 +16,12 @@ rospkg==1.5.0
 tensorflow
 torch==1.13.1
 tqdm==4.64.0
-transformers[sentencepiece]==4.26.0
\ No newline at end of file
+transformers[sentencepiece]==4.26.0
+catkin_pkg==0.5.2
```
Juphex, line 20: could you please sort them as we had in earlier commits? :)
furthermore:
empy and rospkg are already in the requirements, so there is no need to add them

yunlong-wang-cn:
@Juphex thanks for your review.  I want to merge speech_node and command_node together, what is your opinion?

marimbabot_command_recognition/src/__init__.py
yunlong-wang-cn: done

marimbabot_command_recognition/src/text2speech_client.py
yunlong-wang-cn: done

requirements.txt
yunlong-wang-cn: done

Juphex:
@yunlong-wang-cn I did not take a deeper look onto it yet and am not quite sure. Does it make sense for you? Would the project benefit for it in terms of e.g. maintainability? I assume that there would be more cohesion if you would stick to splitting nodes. Please tell me whether you agree/disagree. It is up to you

marimbabot_command_recognition/msg/Command.msg
```diff
@@ -0,0 +1,8 @@
+
```
Juphex, line 1: We may move the messages into our marimbabot_msgs package as well, shouldn't we?

yunlong-wang-cn: Yeah, we can. From my side, it's easy to change and won't cause any problems. Just take care don't have the same name of an ambiguous name of msg files.

Flova:
Why did we switch from a working neural tts (mimic that runs locally)  to espeak (which is fast but sounds bad) to an online neural solution?

marimbabot_command_recognition/msg/Command.msg
yunlong-wang-cn: please just merge this PR, and then I will create a patch to rebase the msg.

marimbabot_speech/config/commands.txt
yunlong-wang-cn, line 1: please tell me if you want any of command

yunlong-wang-cn:
This PR is prepared to be merged if no more questions, the performance is good enough, thanks Florian for providing a nice solution about adding a prompt.

marimbabot_speech/src/text2speech.py
```diff
@@ -35,28 +31,11 @@ def callback_response(self, response):
...
-            #     mimic_subprocess.wait()
-
-            # sound_request.arg = audio_filename
-
         # publish audio file to audio node (sound_play package)
         self.audio_pub.publish(sound_request)
 
 if __name__ == '__main__':
-    rospy.init_node('speech_synthesis_node')
+    rospy.init_node('speech_stt_node')
```
Juphex, line 38: stt=speechtotext? maybe rename the node

yunlong-wang-cn: yeah, actually it is a standard abbrevation, it is commonly used. perhaps I can declarify it somewhere.

marimbabot_speech/msg/Command.msg
```diff
@@ -0,0 +1,4 @@
+std_msgs/Header header
+string command
```
Flova, line 2: Please move the msgs to the msgs package

marimbabot_speech/package.xml
```diff
@@ -3,20 +3,30 @@
...
   <version>0.0.0</version>
   <description>
-  The marimbabot_speech package is designed produce audio from a given text string, acting as a text to speech synthesis.
+  The marimbabot_speech package is designed to recognize and extract the command from the human voice, and also speech synthesis.
+    It consists of the following nodes:
+    - audio_capture: capture audio from microphone (from ros audio_common package)
+    - speech_tmp_node: convert audio to temporary file
+    - speech_stt_node: recognize speech from audio file
+    - speech_ttc_node: convert text to command
+    - speech_tts_node: convert text to speech
```
Flova, line 12: No tts anymore

Flova (approved):

marimbabot_speech/package.xml
yunlong-wang-cn: for TTS, all the codes are from `speech_systhesis.py`, I just renamed it. Is that acceptable?

yunlong-wang-cn: @Flova  If you are ok with the rename, then I will merge it.

PR #147, opened 2023-06-14
Make sides of marimba model smaller

Juphex:
set side with to 0.028


yunlong-wang-cn (approved):
fine for me. good to go.

Flova (approved):

PR #151, opened 2023-06-15
Fix synthesis node

scaredycode:
small changes to the synthesis node

### DIFF
--- a/marimbabot_speech/src/speech_synthesis.py
+++ b/marimbabot_speech/src/speech_synthesis.py
@@ -10,7 +10,7 @@ class SpeechSynthesis:
     def __init__(self):
         # initialize voice and speed
         self.voice = "en_UK/apope_low"
-        self.speed = "1.0"
+        self.speed = "1.5"
 
         # listens to the defined text responses from behavior node
         self.response_sub = rospy.Subscriber("behavior_node/response", String, self.callback_response)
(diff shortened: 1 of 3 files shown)


ateRstones (approved):
Looks good to me, worked on simple inputs without a problem

Juphex (approved):
What is the plan for the speech_synthesis node? Stay on sound_play or switch to e.g. gtts ?

yunlong-wang-cn:
@scaredycode  Sorry for not being clear.  I merged the speech synthesis and  command recognition together, and haven't changed any of your code(except your commented code, just delete it)

yunlong-wang-cn:
> What is the plan for the speech_synthesis node? Stay on sound_play or switch to e.g. gtts ?

@Juphex  As discussed, I think what we to do is merge those two nodes, but don't change any from the speech synthesis that we already have in the main branch.

marimbabot_speech/src/speech_synthesis.py
yunlong-wang-cn, line 1: why did you change it back to the mimic3?

yunlong-wang-cn (changes requested):

scaredycode:
> What is the plan for the speech_synthesis node? Stay on sound_play or switch to e.g. gtts ?

Right now we stay on sound_play. Gtts would only replace mimic/espeak, not the spund_play node.
We can decide wether we want to switch from mimic3 to gtts, but at the moment mimic works just fine and works locally which is a big plus in my opinion.

Yesterday, Michael also proposed a sound_play alternative called audio_play which also is a ROS node, but instead of audio files it only takes the audio content as input. This is nice because then we no longer need tempfiles but keep the information about the audio such as expected playing duration. However, audio_play lacks proper documentation and it might be impossible/difficult to stop audio from playing (which is easily done with sound_play).

scaredycode:
> @scaredycode  Sorry for not being clear.  I will merge the speech synthesis node and command node together, and keep your code unchanged.
> 
> btw, why did you change back to the mimic?

It's what we decided yesterday. Mimic3 just sounds alot better than espeak. Apparently people really were'nt happy I replaced it with the espeak in the first place 😅

scaredycode:
> > What is the plan for the speech_synthesis node? Stay on sound_play or switch to e.g. gtts ?
> 
> @Juphex  As discussed, I think what we to do is merge those two nodes, but don't change any from the speech synthesis that we already have in the main branch.

Now I am confused a bit. I thought yesterday we decided on a different approach. We discussed that the command node will send it's error/status stuff to the behavior node for it to decide on a proper response to send to the speech node. Since the speech node is used in multiple scenarios other than the command detection context (such as status/error in the planning, vision, etc.) I don't completely agree that having the speech synthesis inside the command detection node is the way to go. But if it makes sense to you then I'm also fine with doing it anyways. Maybe @Flova @Juphex have an opinion on this?

marimbabot_speech/src/speech_synthesis.py
scaredycode: It's what we decided yesterday. Mimic3 just sounds alot better than espeak. Apparently people really were'nt happy I replaced it with the espeak in the first place 😅

yunlong-wang-cn:
> > As discussed, I think what we to do is merge those two nodes, but don't change any from the speech synthesis that we already have in the main branch.
> 
> @yunlong-wang-cn I guess merging both modules is fine. I just want to point out, as we discussed, that the STT module will still need to send it's error/status stuff to the behavior node first for it to decide on a proper response to send to the TTS module.



> > As discussed, I think what we to do is merge those two nodes, but don't change any from the speech synthesis that we already have in the main branch.
> 
> @yunlong-wang-cn I guess merging both modules is fine. I just want to point out, as we discussed, that the STT module will still need to send it's error/status stuff to the behavior node first for it to decide on a proper response to send to the TTS module.

I think you mean the command node?  The command node will publish all the recognized commands directly to a topic(/speech_node/command), and all the interaction behavior(e.g. speak something) will be totally decided by the behavior node. Is that clear for you?

yunlong-wang-cn (approved):

Flova (approved):

PR #156, opened 2023-06-21
lilypond to audio behavior integration

Juphex:
I removed the node that listens to the vision node by default.

I changed it so it is an action and may fit better with our behavior.
Please let me know if that fits for you, I also implemented something inside the behavior files fyi @scaredycode .

### DIFF
--- a/marimbabot_audio/src/audio_from_lilypond.py
+++ b/marimbabot_audio/src/audio_from_lilypond.py
@@ -84,7 +84,9 @@ def listener():
 
     # subscribe to vision node and publish to audio node (sound_play package)
     pub = rospy.Publisher('robotsound', SoundRequest, queue_size=50)
-    rospy.Subscriber("vision_node/recognized_notes", String, callback_vision_results, callback_args=(pub))
+
+    # Especially for the behavior node: Topic to subscribe to: lilypond_to_audio.
+    rospy.Subscriber("lilypond_to_audio", String, callback_vision_results, callback_args=(pub))
 
     # spin() simply keeps python from exiting until this node is stopped
     rospy.spin()
--- a/marimbabot_audio/src/audio_from_lilypond.py
+++ b/marimbabot_audio/src/audio_from_lilypond.py
@@ -1,7 +1,10 @@
 #!/usr/bin/env python
 
 import tempfile
+import time
+import wave
 
+import actionlib
 import rospy
 from abjad import Block, LilyPondFile, Score, Staff, Voice
 from abjad.persist import as_midi
@@ -9,6 +12,16 @@ from midi2audio import FluidSynth
 from sound_play.msg import SoundRequest
 from std_msgs.msg import String
 
+from marimbabot_msgs import (LilypondAudioAction, LilypondAudioFeedback,
+                             LilypondAudioResult)
+
+
+def get_audio_length(file_path):
+    with wave.open(file_path, 'rb') as audio_file:
+        framerate = audio_file.getframerate()
+        frames = audio_file.getnframes()
+        duration = frames / float(framerate)
+        return duration
 
 # create lilypond file from sentence
 def create_lilypond_file(sentence):
@@ -58,7 +71,7 @@ def create_audio_from_lilypond(sentence):
     return audio_filename
 
 # callback for vision node, includes audio publisher
-def callback_vision_results(data: String, audio_publisher):
+def callback_lilypond_sentence(data: String, audio_publisher):
     rospy.logdebug("received recognized sentence")
     
     # create audio from lilypond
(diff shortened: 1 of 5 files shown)


Juphex:
did not test it on my machine yet

scaredycode (approved):
it works :)

scaredycode:
action does not "close" properly. When calling the action server again, it is busy everytime.

marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -76,19 +89,59 @@ def callback_vision_results(data: String, audio_publisher):
...
+            durations_seconds = get_audio_length(audio_filename)
+            time.sleep(durations_seconds)
+            self._feedback.in_progress = False
+        except:
+            # if an error happens during audio generation, set success to false
+            success = False
+            rospy.logerr("Error during audio generation inside audio_from_lilypond action server")
+
+        self._result.success = success
+        if success:
```
Flova, line 135: Add an else case for failures. And you might want to set `in_progress` to `false`

Flova: This might solve @scaredycode 's issue

marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -76,19 +89,59 @@ def callback_vision_results(data: String, audio_publisher):
...
+class AudioFromLilypondActionServer(object):
+    # create messages that are used to publish feedback/result
+    _feedback = LilypondAudioFeedback()
+    _result = LilypondAudioResult()
+
+    def __init__(self, name):
+        # publish to audio node (sound_play package)
+        self.pub = rospy.Publisher('robotsound', SoundRequest, queue_size=50)
+
+        self._action_name = name
```
Flova, line 106: The node and the action have the same name. Idk if I like this.

marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -1,14 +1,27 @@
...
 from abjad.persist import as_midi
 from midi2audio import FluidSynth
 from sound_play.msg import SoundRequest
 from std_msgs.msg import String
 
+from marimbabot_msgs.msg import (LilypondAudioAction, LilypondAudioFeedback,
+                             LilypondAudioResult)
+
+
+def get_audio_length(file_path):
```
Flova, line 19: Why are these methods separate and not part of the node class?

marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -58,7 +71,7 @@ def create_audio_from_lilypond(sentence):
     return audio_filename
 
 # callback for vision node, includes audio publisher
-def callback_vision_results(data: String, audio_publisher):
+def callback_lilypond_sentence(data: String, audio_publisher):
```
Flova, line 74: This isn't a callback anymore so consider renaming it.

marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -76,19 +89,59 @@ def callback_vision_results(data: String, audio_publisher):
...
 
-# create audio from lilypond publisher
-# send audio to audio node (sound_play package)
-def listener():
+    return audio_filename
+
+"""
+Action server for audio generation from lilypond string
+"""
+class AudioFromLilypondActionServer(object):
```
Flova, line 97: Idk if I would name it action server. You also don't name the classes of other nodes `XYZSubscriber`. It is not wrong, but the action server is just an interface for this code.

Juphex: done

marimbabot_audio/src/audio_from_lilypond.py
Juphex: done

marimbabot_audio/src/audio_from_lilypond.py
Juphex: done

marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -76,19 +89,59 @@ def callback_vision_results(data: String, audio_publisher):
...
+
+        self._action_name = name
+        self._as = actionlib.SimpleActionServer(self._action_name, LilypondAudioAction, execute_cb=self.execute_cb, auto_start = False)
+        self._as.start()
+      
+    def execute_cb(self, goal):
+        # helper variables
+        success = True
+        
+        # append the seeds for the fibonacci sequence
```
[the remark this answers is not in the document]
Juphex: done

marimbabot_audio/src/audio_from_lilypond.py
Juphex: I added in_progress and an else case.
I could not replicate the problem of @scaredycode   where the action server cannot be called multiple times / is busy after calling it again.

Flova (approved):
Looks good to me, but please test it with the rest of the system like last week with @scaredycode .

Juphex:
@scaredycode @Flova 
I could reproduce the error "Actioin server is busy." now.
The fix would be re-initializing the action client every time before sending the notes to it.
I.e. `        self.lilypond_audio_client = actionlib.SimpleActionClient('audio_from_lilypond', LilypondAudioAction)`

This works as intended but I don't know why it does not work without re-initializing the client. I mean the action should be busy independent from a fresh-initialized client object.

scaredycode:
> @scaredycode @Flova I could reproduce the error "Actioin server is busy." now. The fix would be re-initializing the action client every time before sending the notes to it. I.e. ` self.lilypond_audio_client = actionlib.SimpleActionClient('audio_from_lilypond', LilypondAudioAction)`

Flo and I found the issue. The way I checked if the action is busy in the behavior was not valid. The action works just fine. 
We tested the fix and now everything works as intended. I committed the behavior fix in this PR, hope you don't mind.

Juphex:
amazing 😊 @Flova  @scaredycode

PR #158, opened 2023-06-22
Update marimbabot.launch

yunlong-wang-cn:
change the name of node to avoid conflict


yunlong-wang-cn:
@scaredycode  thanks for finding the conflict, and you can take a look.

yunlong-wang-cn:
> It also makes sense for consistency reasons.
> Would you like to change the other launch files as well ? 

I will change also the node name for feature/audio. Perhaps in the audio_patch PR, just not here.

PR #161, opened 2023-06-29
fix package.xml

yunlong-wang-cn:
fix bug from #160
@scaredycode please check it. 
Or you can try to use `rosdep install --from-paths src --ignore-src -r -y` to install the dependecies.

### DIFF
--- a/marimbabot_audio/package.xml
+++ b/marimbabot_audio/package.xml
@@ -22,7 +22,6 @@
   <build_depend>sound_play</build_depend>
 
   <exec_depend>message_runtime</exec_depend>
-  <exec_depend>audio_common</exec_depend>
   <exec_depend>fluidsynth</exec_depend>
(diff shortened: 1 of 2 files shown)


yunlong-wang-cn:
@scaredycode  If you have time, pls take a look if this problem still exists.

Juphex (commented):
I assume the "bug" is just adding audio_capture to package.xml, right @yunlong-wang-cn ?

lgtm if so

yunlong-wang-cn:
> I assume the "bug" is just adding audio_capture to package.xml, right @yunlong-wang-cn ?
> 
> lgtm if so

@Juphex  You are right, but I haven't met this issue with my computer, so better let @scaredycode check if it is resolved.

Flova (approved):

scaredycode (approved):

PR #166, opened 2023-06-29
include command message

scaredycode:
@yunlong-wang-cn would be cool if you could test it for me :)


yunlong-wang-cn (approved):
It looks pretty good.

PR #176, opened 2023-07-06
two mallet configuraitons for motion

berkgungor:
- two mallet configurations and new launch file based on mallet selection
- joint constraints for more stable trajectory
- dummy joint state publisher


marimbabot_hardware/src/joint_state.py
Juphex, line 1: why does this file occur in /scripts and /src?

marimbabot_hardware/src/joint_state.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
+
+import rospy
+from sensor_msgs.msg import JointState
+
+def JointStatePublisher():
+    pub = rospy.Publisher('/joint_states', JointState, queue_size=10)
+    rospy.init_node('joint_state_publisher')
+    rate = rospy.Rate(100) # 10hz
```
Juphex, line 9: should be 100hz then :)

marimbabot_planning/src/planning.cpp
```diff
@@ -85,17 +85,17 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::plan_to_mallet_po
...
         goal_point.point.x,
         goal_point.point.y,
-        goal_point.point.z);
+        goal_point.point.z + 0.1);
 
     // Use bio_ik to solve the inverse kinematics at the goal point
     bio_ik::BioIKKinematicsQueryOptions ik_options;
     ik_options.replace = true;
-    ik_options.return_approximate_solution = false; // Activate for debugging if you get an error 
+    ik_options.return_approximate_solution = true; // Activate for debugging if you get an error
```
Juphex, line 93: Do you want to change back to false it before merging?

marimbabot_vision/scripts/generate_data.py
```diff
@@ -1,11 +1,12 @@
 #!/usr/bin/env python3
```
Juphex, line 1: where do these changes come from?

Juphex (changes requested):

marimbabot_planning/src/planning.cpp
berkgungor: no harm to keep it for now i think

Juphex (approved):
running marimbabot_ur5_flex_double_moveit_config)/launch/demo.launch works
event two mallet planning

Even when I add a goal for each mallet_head planning works

marimbabot_planning/src/planning.cpp
```diff
@@ -110,7 +110,21 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::plan_to_mallet_po
...
+        tf2::Quaternion desired_orientation;
+        desired_orientation.setRPY(0.52, 0.0, 0.0); // Set roll, pitch, and yaw angles
+
+        // Calculate the angular distance between the current and desired orientations
+        tf2::Quaternion orientation_error = desired_orientation.inverse() * orientation;
+        double angular_distance = 1.0 - orientation_error.dot(orientation_error);
+        return std::pow(angular_distance, 2);
+    };
+
+    ik_options.goals.emplace_back(new bio_ik::LinkFunctionGoal("ur5_wrist_2_link", orientation_constraint));
```
Flova, line 126: Why did you use the LinkFunctionalGoal, which is in cartesian space instead of a normal JointGoal for the over actuated wrist joint?

Flova: @berkgungor

marimbabot_planning/src/planning.cpp
```diff
@@ -222,8 +236,8 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
 {
     try {
         // Set the max velocity and acceleration scaling factors
-        move_group_interface_.setMaxVelocityScalingFactor(0.9);
-        move_group_interface_.setMaxAccelerationScalingFactor(0.9);
+        move_group_interface_.setMaxVelocityScalingFactor(0.5);
```
Flova, line 239: Why did you lower the speed?

marimbabot_planning/src/planning.cpp
```diff
@@ -85,17 +85,17 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::plan_to_mallet_po
     tf2::Vector3 goal_position(
         goal_point.point.x,
         goal_point.point.y,
-        goal_point.point.z);
+        goal_point.point.z + 0.1);
```
Flova, line 88: Offsetting the goals and changing the physical position of the marimba is a dirty hack to avoid collisions, which should not be done normally.

marimbabot_planning/src/planning.cpp
```diff
@@ -85,17 +85,17 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::plan_to_mallet_po
...
 
     // Use bio_ik to solve the inverse kinematics at the goal point
     bio_ik::BioIKKinematicsQueryOptions ik_options;
     ik_options.replace = true;
-    ik_options.return_approximate_solution = false; // Activate for debugging if you get an error 
+    ik_options.return_approximate_solution = true; // Activate for debugging if you get an error 
 
-    ik_options.goals.emplace_back(new bio_ik::PositionGoal("mallet_head", goal_position));
+    ik_options.goals.emplace_back(new bio_ik::PositionGoal("mallet_head_1", goal_position));
```
Flova, line 95: Maybe this should be configurable, so we still support the single mallet case.

PR #185, opened 2023-07-09
increase volume

Juphex:


marimbabot_audio/src/audio_from_lilypond.py
```diff
@@ -131,7 +131,7 @@ def create_and_publish_audio(self, data: String, audio_publisher):
         sound_request = SoundRequest()
         sound_request.sound = -2
         sound_request.command = 1
-        sound_request.volume = 3
+        sound_request.volume = 5.0
```
scaredycode, line 134: While it is definitely louder than before, I feel like you could go even louder. It is still a bit quiet in comparison to the speech synthesis. I played around with different values and imo 10 is pretty good.

scaredycode (changes requested):

marimbabot_audio/src/audio_from_lilypond.py
Juphex: You are right. 
I also researched whether the python package we use (midi2audio) allows us to use a fluidsynth but I did not find anything. So, I changed the sound_request volume to 10.0. :)

PR #186, opened 2023-07-09
communicate exact time of the first bar hit for audio feedback

Juphex:
I adjusted the Action result definition, so that timings are returned with the same message type as the goal is defined (List of HitSequenceElements).
These HitSequenceElements are returned with an updated start_time value.
What I do is that I get the duration from start of the planned trajectory (http://docs.ros.org/en/noetic/api/trajectory_msgs/html/msg/JointTrajectoryPoint.html ) and add this to the time when the plan is actually executed.

Does that make sense for you @yunlong-wang-cn ? :)
Here are two outputs from the returned list.
The first half is for the sequence "C" and second half is for "C,D,E".
You can see that I printed the new "executed_sequence_elements" list directly.

image


scaredycode (approved):
LGTM

scaredycode:
although "start of trajectory" might not necessarily (or most likely not) be the time the first note is hit (because the robot must first move to the first bar which takes extra time)
I would prefer the action to give that time stamp (planned contact with first bar) as feedback instead, so that @yunlong-wang-cn gets this information beforehand

yunlong-wang-cn:
yes,  and I want to make sure my understanding is right.
The following is the content of HitElement.msg:
  
    string tone_name
    int32 octave
    time start_time
    duration tone_duration
    float32 loudness

So, the `start_time` is still the relative time that starts from zero, and therefore the `duration` is given by the visual node, not from the motion planning, therefore it could be not matched with the `start_time`.

@scaredycode In a word, I should use `start_time` to compare the sequence, right?

Juphex:
> yes, and I want to make sure my understanding is right. The following is the content of HitElement.msg:
> 
> ```
> string tone_name
> int32 octave
> time start_time
> duration tone_duration
> float32 loudness
> ```
> 
> So, the `start_time` is still the relative time that starts from zero, and therefore the `duration` is given by the visual node, not from the motion planning, therefore it could be not matched with the `start_time`.
> 
> @scaredycode In a word, I should use `start_time` to compare the sequence, right?

It is quite fiddly, but this result sends start_time in the form of **absolute time**. Not relative :)

yunlong-wang-cn:
file_720
As discussed with @Juphex , 

- Step 1: behavior_node will send the primary plan to the motion_planning_node after the vision_node's recognization.
- Step 2: motion_planning_node will feedback the modified plan or actual plan to behavior_node
- Step 3: The behavior_node will send the final plan to the audio node, **there the `start_time` should be absolute time.**

Juphex:
I am going to test it with a dummy at home, then I will merge it.

marimbabot_planning/include/marimbabot_planning/planning.h
```diff
@@ -15,7 +15,7 @@
 #include <tf2_ros/transform_listener.h>
 #include <tf2/LinearMath/Quaternion.h>
 #include <trajectory_msgs/JointTrajectoryPoint.h>
-
+#include <list>
```
Flova, line 18: Please only use a list if you perform random insertions or deletions of elements at other places than the end. In most other cases (like this one) a `std::vector` is the data structure you want to use in cpp.

marimbabot_planning/include/marimbabot_planning/planning.h
```diff
@@ -31,6 +31,10 @@ class Planning
         // MoveIt! MoveGroupInterface
         moveit::planning_interface::MoveGroupInterface move_group_interface_;
 
+        // Describes when a "hit" was performed in terms of 
+        // how long after the start of the trajectory this point should be reached 
+        std::list<ros::Duration> hit_plan_durations;
```
Flova, line 36: Doing this as a side effect of a program that is otherwise mostly [pure](https://en.wikipedia.org/wiki/Pure_function) can lead to issues, as nobody expects that the trajectory needs to be generated in order. One could e.g. generate the approach to the initial note in the end.

marimbabot_planning/src/planning.cpp
```diff
@@ -260,6 +264,14 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::hit_notes(
  */
 void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalConstPtr &goal)
 {
+    // clear the hit plan durations
+    this->hit_plan_durations.clear();
+
+    // initialize feedback
+    marimbabot_msgs::HitSequenceFeedback feedback;
+    feedback.playing = true;
+    feedback.executed_sequence_elements.clear();
```
Flova, line 273: You don't need to clear the freshly initialized vector.

marimbabot_planning/src/planning.cpp
```diff
@@ -291,8 +303,38 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
         display_trajectory.trajectory.push_back(trajectory);
         trajectory_publisher_.publish(display_trajectory);
 
+
+        // set the start time of execution
+        ros::Time start_time = ros::Time::now();
+
+        // set the feedback based on the hit plan
```
Flova, line 310: The could be moved to its own function in the utils.

marimbabot_planning/src/planning.cpp
```diff
@@ -303,6 +345,16 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
             action_server_.setAborted(result);
         } else {
             marimbabot_msgs::HitSequenceResult result;
+
+            // Check if the size of the hit_sequence_elements and the hit_plan_durations was the same
+            // If not, we cannot set the executed_sequence_elements for the result 
+            // and success is kept to false
+            if (goal->hit_sequence_elements.size() != hit_plan_durations.size()) {
```
Flova, line 352: Why would this be the case if not for side effects?

marimbabot_msgs/action/HitSequence.action
```diff
@@ -15,3 +16,4 @@ uint16 error_code # error code in case of failure (see above)
 #feedback
 ---
 bool playing # whether marimbabot is currently (still) playing, i.e. the goal is still active
+marimbabot_msgs/HitSequenceElement[] executed_sequence_elements # see above
```
Flova, line 19: Why is this named in past tense? When it is first published, they are not executed.

marimbabot_planning/src/planning.cpp
```diff
@@ -215,7 +215,10 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::hit_note(
     
     // Concatinate trajectories
     auto plan = concatinated_plan({approach_plan, down_plan, retreat_plan});
-
+    
+    ros::Duration approach_duration = down_plan.trajectory_.joint_trajectory.points.back().time_from_start;
```
Flova, line 219: Why do you only consider the time we need to approach the note, but note the time we need to hit it? This creates an error that adds up. Why do you sum up these durations either way? You only need to determine the time at which we reach the first note and add this to the relative timings provided in our goal. This gets weird if we consider clipping during the trajectory generation. If we include clipped timings here they will be used as ground truth and not considered as an issue, but if we don't all subsequent notes are slightly offset and get low scores. This being said, ideally we should never need to clip timings as they result in "wrong" but feasible timings. Ideally they should result in cancelation of the plan with an appropriate error code. Currently, the initial approach is always clipped, as we can not be at the first note in 0 seconds, but we would need to explicitly handle it in this case.

marimbabot_planning/src/planning.cpp
```diff
@@ -291,8 +303,38 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
...
+
+        // Publish the feedback
+        action_server_.publishFeedback(feedback);
+
         // Execute the plan
         auto status = move_group_interface_.execute(hit_plan);
+        // Set playing to false
+        feedback.playing = false;
+
+        ros::Duration plan_execution_time = ros::Time::now() - start_time;
```
Flova, line 337: This is never used

marimbabot_planning/src/planning.cpp
```diff
@@ -291,8 +303,38 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
...
+            ++duration_iter;
+        }
+
+        // Publish the feedback
+        action_server_.publishFeedback(feedback);
+
         // Execute the plan
         auto status = move_group_interface_.execute(hit_plan);
+        // Set playing to false
+        feedback.playing = false;
```
Flova, line 335: This is never published

Flova (commented):
Sorry for looking at this so late, after it was merged. While looking at it, I noticed quite a lot of issues that still need to be addressed imo.

I think we should rework it so that we only estimate the approach time for the first note and communicate that. We can still use the same message layout or just send the timing and let the behavior combine it with the original goal hit points.

@Juphex @yunlong-wang-cn  @scaredycode

PR #191, opened 2023-07-13
rm unnecessary loc

Juphex:
I guess this is just an artifact of some debugging


scaredycode (approved):
You are right, good catch 👍

PR #193, opened 2023-07-13
add functions for volume modification

Juphex:
@scaredycode 
I implemented some further functions on volume modification.
you may have a look :)

### DIFF
--- a/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
+++ b/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
@@ -42,6 +42,22 @@ class ActionDecider:
             rospy.logwarn('Lilypond string not valid.')
             self.response_pub.publish('Lilypond string not valid.')
 
+    def increase_volume(self):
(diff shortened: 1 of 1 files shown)


scaredycode:
Thanks for helping out with the behavior :)
The idea is nice, but I actually would prefer to alter the lilypond sequence instead. Here's why:

While modifying the hit sequence directly works for planning, the Hit sequence is now different from the lilypond sequence. Therefore, while the planning gets the right instructions, the preview (working with lilypond) will stay the same. Also, giving the command to play faster/slower after modifying volume will now get rid of the volume change as the tempo functions will only modify the lilypond sequence and then regenerate the Hit sequence to replace the current one.

My idea was to Look at the lilypond string and replace every \f, \mp, \p and so on with one symbol louder/quieter, and then just let the Interpreter do its thing. That should ensure tjat the lilypond sequence and the Hit sequence are equal again. Let me know what you think :)

Juphex:
I also thought about your idea and also thought it might be a bit messy when changing it again. :smile: But you can take this as an idea if you want. Keep me up do date about your decision if you like, I will close this PR.

Issue #140, opened 2023-06-08
Create and test a moveit config for the double mallet


ateRstones:
https://github.com/UHHRobotics22-23/marimbabot/tree/feature/two_mallet_demo_mode

Issue #61, opened 2023-04-13
Speed up or slow down trajectories to match loudness


Flova:
Related to #49, but focussed of the hitting part of the trajectory instead of the transition.

Juphex:
also keep in mind that, also given the dynamics in a music sheet, we should be able to do this note/bar-specific

berkgungor:
task is done and merged in main already.

Issue #49, opened 2023-04-13
Speed up or slow down trajectories to match rythm


scaredycode:

ateRstones:
I see this and #61 are connected to the closed pull request #118 . Are these obsolete tickets @berkgungor ?

Juphex:
@ateRstones https://github.com/UHHRobotics22-23/marimbabot/issues/61 is already tackeled. I assume that here it is about changing the speed according to a user input?

berkgungor:
#61 was an old ticket and related to #118 pull request as @ateRstones mentioned and I just closed the ticket #61.  I am also closing this ticket too as it addresses the same issue.

PR #208, opened 2023-08-10
Feature/double mallet loudness

berkgungor:
Speed up and slow down trajectories with timing to catch the rythm.

- Can be tested with timing.py giving loudness, octave and notes as argument.
- Slow down motion and interpolation are added into motion planning.

#49 
#61


ateRstones (approved):
LGTM, a bit convoluted with the contents and I think it also closes #140

Flova (approved):

Issue #213, opened 2023-08-10
Create Rosbags for Integrated or Semiintegrated environments of the project to be used for testing nodes in a isolated state


ateRstones:

yunlong-wang-cn:
I will try to create two rosbag:
1. to collect the audio information from the robot playing with the single mallet, here I need help from @Flova or @berkgungor on how to use the robot, I need to know the interface, if you have it done please let me know.
2. to collect some human speech under common situations (i.e. with some noise and/or robots speaking) to test/calibrate the robustness of the speech recognition.

Juphex:
Can you give an update?
@yunlong-wang-cn

Issue #201, opened 2023-07-20
[marimbabot_audio]: enhancement


yunlong-wang-cn:
1. the midi updates alone the changes of time.
2. to give the notes which mismatched, in the way of midi visualization, e.g show the plan and feedback sequence with two different colors.

yunlong-wang-cn:
done in PR #218, the close

PR #218, opened 2023-08-10
Audio patch2

yunlong-wang-cn:
Following task plan to do in this PR:
#213 (suspended)
#201 (both task in this issue is done)
#270 todo


marimbabot_audio/src/sequence_evaluation.py
```diff
@@ -0,0 +1,161 @@
+import rospy
+from marimbabot_msgs.msg import HitSequence as HitSequenceMsg
+from marimbabot_msgs.msg import SequenceMatchResult as SequenceMatchResultMsg
+from marimbabot_msgs.msg import NoteOnset as NoteOnsetMsg
+
+class Comparison():
```
Flova, line 6: Would it make sense to compare them with dynamic time warping as an extra metric, to filter out constant offsets. Because it would be preferable, in many scenarios, to keep playing with accurate relative timings instead of playing faster to match the original speed again after we missed one timing.

Flova:
What is the state of this PR?

yunlong-wang-cn:
almost done. Just some tuning of the onset detection.

yunlong-wang-cn:
some parameters of the onset detection can be tunned to improve performance, but let we test if it enough now.

marimbabot_audio/src/sequence_evaluation.py
yunlong-wang-cn: how about talking about the details in person, I am not sure what you want to do exactly. And why do you need to compare it with relative time? I will be in the student lab every day.

Flova:
What's the state of this?

yunlong-wang-cn:
@Flova  prepared to be reviewed. Further parameter tunning will be done later.

Flova:
Should it be merged before #230 ?

yunlong-wang-cn:
> Should it be merged before #230 ?

There are no dependencies between #230 and #218.

Juphex:
This branch has recent changes. Is it ready for review?
@yunlong-wang-cn

marimbabot_audio/src/sequence_evaluation.py
Flova: If you compare the absolute timings and the robot is too slow at one note it carries the offset and gets penalized for all subsequent notes. With relative timings you only compare the timing from one note to the other meaning it only gets punished for the one wrong note. 

I don't know if this chart helps. Hits are denoted with `|` and misses are marked with `x`. Each character is a timestamp.
```
Target:      |---|---|----|
Measurement: |----|---|----|
Absolute:    |----x---x----x
Relative:    |----x---|----|
```

yunlong-wang-cn:
@Juphex yes, you can review it.

Juphex (commented):
@yunlong-wang-cn and I talked about this and it looks fine.
Smaller things like hierarchical launch files and sorting imports are good to have.
It seems to be working. Yunlong will also add some README documentation on the launch files.

First, we should merge this branch locally on the robot's computer and try to integrate it before merging it in the origin/main.
And please as a squashed merge 😅

marimbabot_audio/launch/marimbabot_rviz.launch
```diff
@@ -1,5 +1,8 @@
 <?xml version="1.0"?>
 <launch>
-    <include file="$(find marimbabot_audio)/launch/marimbabot.launch"/>
+    <include file="$(find marimbabot_audio)/launch/audio_feedback.launch"/>
```
Juphex, line 3: Please revert to:

```
<?xml version="1.0"?>
<launch>
    <include file="$(find marimbabot_audio)/launch/marimbabot.launch"/>
    <node name="audio_node_rviz" pkg="rviz" type="rviz" args="-d $(find marimbabot_audio)/config/config.rviz"/>
</launch>
```

Juphex:
Please remove
`marimbabot_audio/test/.ipynb_checkpoints/onset_detection-checkpoint.ipynb`

marimbabot_audio/launch/marimbabot_rviz.launch
yunlong-wang-cn: I deleted this file, since start the rviz from the launch may not receive the message correctly.

Flova (approved):

Flova (approved):

PR #224, opened 2023-08-14
211 adjust the synthetic dataset generation

Juphex:
This PR fixes the dotted note problem (too many clocks in a bar)
simplifies the data that is generated by default
Adds a rotation to the output image, so this and the handwritten data are in portrait mode. (make generation a bit slower because it adds an extra read and write step)


Juphex:
Dotted note examples:

grafik
grafik

scaredycode:
The new beats-per-bar system works perfectly :)
However, some samples have 16th notes even though I set min_duration to 8. You might wan't to have a look at that. 
staff_1

Juphex:
now the dotted note is not always followed by its duration/2

grafik

Flova (approved):

Flova:
Lets also wait for the review of @scaredycode

scaredycode (changes requested):
Please remove the double sharps & double flats. Also would be nice to randomize dynamics for every note (or every beginning bar) instead of just at the first position. Otherwise looks good.

PR #226, opened 2023-08-16
Use the flex double mallet in simulation

### LINKED ISSUE #227: Make simulation work with double mallet holder

iibrahimli:
Switch the mallet holder type to `flex_double` in simulation and make appropriate changes, namely:
- Add controllers to double mallet moveit config (copied as-is from old moveit config)
- Switch to double mallet moveit config (i.e. use launch files from the new config)
- Remove LimitMaxCartesianLinkSpeed (not needed, its non-existence was causing an error)

Closes #227


marimbabot_ur5_flex_double_moveit_config/config/joint_state_controller.yaml
iibrahimli, line 1: Also copied over from `marimbabot_ur5_moveit_config/config`

berkgungor (approved):
LGTM

PR #231, opened 2023-08-16
forward message of planned hit sequences to topic audio/hit_sequence

Juphex:
this message type contains a list of HitSequenceElement

@yunlong-wang-cn We should talk about whether this fits you (receiving a list of notes all the time) or if you prefer to receive a single note per "publish".


yunlong-wang-cn (approved):
Looks pretty good, you can merge it.

PR #232, opened 2023-08-17
Adjusted the base link offset of the double mallet

ateRstones:


Flova (approved):

Juphex:
Can we merge it or not? @ateRstones

ateRstones:
@Flova I noticed a difference with the mounting "ridge" not aligning corrently. The model was different before, so I had to adjust the offset based on different markers, which was close but not completely correct

PR #238, opened 2023-08-17
correct the typo

yunlong-wang-cn:
correct the typo for setup the hardware.


yunlong-wang-cn:
berk will correct it with his PR, so withdraw this PR

Issue #209, opened 2023-08-10
pattern  of command recognition


yunlong-wang-cn:
@scaredycode 
Consider adding the pattern for the command recognition

yunlong-wang-cn:
@scaredycode Sorry, I can not find which pattern we need to recognize for command recognition, I think you mentioned it somewhere, could you show me again?

scaredycode:
> @scaredycode Sorry, I can not find which pattern we need to recognize for command recognition, I think you mentioned it somewhere, could you show me again?

 I talked about such pattern commands in this issue here:
https://github.com/UHHRobotics22-23/marimbabot/issues/182

yunlong-wang-cn:
@scaredycode  I haven't used the LLM to build the command extraction, it is too huge,  can not be run in the PC near UR5, so I write a much single logic with regular expression, and it work pretty goods with the pattern above.

Flova:
Is this issue also covering robustness against things like `marimbabot plays`, `marimbabot, plays`, ...?

yunlong-wang-cn:
@scaredycode For your PR#242, give you info from this PR#218. 
Here give you the list of all possible commands in the format of `{"action": "increase speed", "parameter": "20bpm"}`.

The following "parameter" is always changeable and could be an empty string to denote user hasn't given any parameters of this action.

1. For speed: 
 `{"behavior": "play", "action": "increase speed", "parameter": "20bpm"}`.
 `{"behavior": "play", "action": "decrease speed", "parameter": "40bpm"}`
 `{"behavior": "play", "action": "setup speed", "parameter": "20bpm"}`
 2. For volume:
  `{"behavior": "play", "action": "increase volume", "parameter": "20bpm"}`.
 `{"behavior": "play", "action": "decrease speed", "parameter": "40bpm"}`
 `{"behavior": "play", "action": "setup speed", "parameter": "20bpm"}`
3. other:
"play" of behavior can be replace by `preview` or `loop`

yunlong-wang-cn:
@scaredycode FYI

Issue #182, opened 2023-07-06
Ideas for additional commands 


scaredycode:
- play in specific tempo -> "Marimbabot play in 120 bpm" - already supported in behavior
- play in specific loudness -> "Marimbabot play in forte"
- increase/decrease Tempo by specific value -> "Marimbabot play faster by 60 bpm" - already supported in behavior
- increase/decrease loudness by specific value -> "Marimbabot play louder by 20 percent" 
- all "play" commands also as "preview" -> "Marimbabot preview", "Marimbabot preview slower by 30 bpm", "Marimbabot preview in 80 bpm", ...  - already supported in behavior
- save piece for future use -> "Marimbabot save as Testsong"
- play from memory -> "Marimbabot play Testsong"

Juphex:
All commands are here https://github.com/UHHRobotics22-23/marimbabot/blob/main/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py, right?

Maybe something like play everything a half-note / note higher? :)

scaredycode:
> All commands are here https://github.com/UHHRobotics22-23/marimbabot/blob/main/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py, right?

What do you mean by "are here" exactly? Currently the behavior can work with the commands that are marked as "already supported in behavior". However, these commands are not supported by the speech module yet, meaning while we theoretically can work with the instruction, we cant recognize it in the speech detection layer.
The issue we face right now is the recognition of not only a given text but a pattern instead ("play in <x> bpm"

yunlong-wang-cn:
new command: **`play/preview in a loop`**

old command:
play in **specific tempo** -> "Marimbabot play in 120 bpm" - already supported in behavior
play in **specific loudness** -> "Marimbabot play in forte"
**increase/decrease Tempo** by specific value -> "Marimbabot play faster by 60 bpm" - already supported in behavior
**increase/decrease loudness** by specific value -> "Marimbabot play louder by 20 percent"
all "play" commands also as **"preview"** -> "Marimbabot preview", "Marimbabot preview slower by 30 bpm", "Marimbabot preview in 80 bpm", ... - already supported in behavior

Juphex:
Can we close this @scaredycode  @yunlong-wang-cn  ?

yunlong-wang-cn:
yes, all command stuff settle down.

PR #242, opened 2023-08-17
Add "louder", "softer" and "loop"

scaredycode:
also some fixes (for tempo assignment)


marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -151,69 +206,71 @@ def callback_command(self, command_msg):
                 rospy.logwarn('No notes recognized.')
                 self.response_pub.publish('No notes recognized.')
 
-        # play the notes on the marimba using the UR5
-        elif command == 'marimbabot start playing':
+        # play the notes on the marimba using the UR5 or generate an audio preview
+        elif command == 'marimbabot start (playing|preview)':
```
Flova, line 210: I don't think this works

Flova: As this is a normal equal operation and not a regex

scaredycode: You are right! Must have missed this. I fixed it now

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -40,6 +40,14 @@ def __init__(self):
         # action client to send the note_sequence to the lilypond_audio action server
         self.lilypond_audio_client = actionlib.SimpleActionClient('audio_from_lilypond', LilypondAudioAction)
 
+        # Waits until the action server has started up and started
+        # while not self.planning_client.wait_for_server(timeout=rospy.Duration(2)) and not rospy.is_shutdown():
+        #     rospy.loginfo('Waiting for the planning action server to come up')
+
+        # Waits until the action server has started up and started
+        while not self.lilypond_audio_client.wait_for_server(timeout=rospy.Duration(2)) and not rospy.is_shutdown():
```
yunlong-wang-cn, line 48: I prefer to setup a timeout is better, just increase the sec to be long enough. but it's fine.

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -151,69 +225,110 @@ def callback_command(self, command_msg):
...
+
+                # if the preview is not busy, the note sequence is send to the lilypond_audio action server
+                if 'preview' in command:
+                    if self.lilypond_audio_client.simple_state == actionlib.SimpleGoalState.DONE:
+                        self.preview(loop)
+                    else:
+                        rospy.logwarn('The audio preview is already playing.')
+                        self.response_pub.publish('The audio preview is already playing.')
             else:
                 rospy.logwarn('No notes to play. Say reed to read notes.')
```
yunlong-wang-cn, line 253: typo `reed`

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -151,69 +225,110 @@ def callback_command(self, command_msg):
...
                 self.response_pub.publish('No notes to play. Say reed to read notes.')
```
yunlong-wang-cn, line 254: typo `reed`

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -151,69 +225,110 @@ def callback_command(self, command_msg):
...
-                    value = int(command.split(' ')[-2])
-                    self.assign_tempo(value)
-                self.preview()
-            else:
-                rospy.logwarn('No notes to preview. Say reed to read notes.')
-                self.response_pub.publish('No notes to preview. Say reed to read notes.')
-
-        # preview faster or slower than the current tempo (by specified bpm value, default = 20)
-        elif re.match(r'marimbabot preview (faster|slower)( by [0-9]+)?', command):
+        elif re.match(r'marimbabot (play|preview) (louder|softer)( by [0-9]+ steps)?', command):
```
yunlong-wang-cn, line 292: For the coming version of the command recognition, the keyword `marimbabot` is not contained. because the STT is triggered after the keyword `marimbabot`.

If you do need it I and add it manually.

yunlong-wang-cn (changes requested):
The most important is, in the coming version of command recognition, the content of the command message is changed, it is now a JSON string inside, like `{"action": "increase speed", "parameter": "20bpm"}`. I have already talked to @scaredycode, and she is good with it. So perhaps you need to change the way to parse the behavior, I think it is much easier to do it.
And there is also another option, that I can open the command recognition to you, which means if you want you can directly parse the command from the recognized text.

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
scaredycode: I fixed the type in the logwarn message. Unfortunately, it has to stay in the response string, as the TTS would pronounce it as past tense otherwise (rĕd instead of rēd)

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
scaredycode: Not sure I understand. Which timeout are you referring to? All this timeout (2s) does, is defining a time interval, in which the message "Waiting for the ... server to come up" is printed over and over again until the server is reached. It just serves as an indication to us if there are issues reaching the servers.
The actual time it takes for the servers to start is not in any way restricted by it. Currently, if a server does not ever start, this message would be printed every 2s until the code is shutdown.

yunlong-wang-cn:
@scaredycode all possible command refer to this issue#182, and all possible json refer to here #209

berkgungor (approved):
LGTM

Flova (approved):

berkgungor:
@scaredycode when we try to send "play louder" command, behaviour node sets "90" as default note speed and increases with +5 or +6 every single time that we send the louder command and since we have 7 loudness option, we can increase the loudness level up to 107 while the maximum value is supposed to be 127. (in prettyMIDI library, notes have a velocity (volume) from 1 to 127). You can follow the issue : #258

berkgungor (changes requested):
@scaredycode when we try to send "play louder" command, behaviour node sets "90" as default note speed and increases with +5 or +6 every single time that we send the louder command and since we have 7 loudness option, we can increase the loudness level up to 107 while the maximum value is supposed to be 127. (in prettyMIDI library, notes have a velocity (volume) from 1 to 127). You can follow the issue : #258

marimbabot_behavior/src/marimbabot_behavior/interpreter.py
```diff
@@ -55,12 +55,12 @@ def read_notes(notes) -> HitSequenceGoal:
 
     for note in midi_stream.instruments[0].notes:
         name = pm.note_number_to_name(note.pitch)
+        normalizer = 40
```
berkgungor, line 58: Tested with normalization value.
Minimum loudness was 0.15 and maximum was 0.67, on the other hand we observe maximum loudness as 1.0 when we pass the "mercato" string.
30 sounded little more louder since we get less loud sound when we increase the normalize value. But 40 also works.

berkgungor (approved):

scaredycode:
> @scaredycode when we try to send "play louder" command, behaviour node sets "90" as default note speed and increases with +5 or +6 every single time that we send the louder command and since we have 7 loudness option, we can increase the loudness level up to 107 while the maximum value is supposed to be 127. (in prettyMIDI library, notes have a velocity (volume) from 1 to 127). You can follow the issue : #258

The "add normalization" commit increases the volume range. The new minimum volume (\ppp) is now at ~ 0.15, while the max valume (\fff - \marcato) remains at 1.0.

marimbabot_behavior/src/marimbabot_behavior/interpreter.py
berkgungor: Good to merge. Tests were successfull

scaredycode:
> @scaredycode all possible command refer to this issue #182, and all possible json refer to here #209

I adapted the existing code to the new command message format. Please have a look :)

yunlong-wang-cn (approved):
LGTM :+1:

PR #244, opened 2023-08-21
237 behavior node dummy client for send plan to motion planning

Juphex:
@yunlong-wang-cn 
As we discussed, here is a dummy node with that you can publish the audio input/commands and the vision input. :+1: 
You should just launch the new launch file and the new node separately.
I still have some bug that the behavior node dies, I try to solve that.
However, you can still give it a try and look whether it works already.
Was I correct that you want to simulate the vision and command inputs?

@scaredycode Do we "need" the other launch file "test_behavior.launch" or was it just for local / one-time-use testing? :)

### DIFF
--- a/marimbabot_behavior/scripts/behavior_node_dummy.py
+++ b/marimbabot_behavior/scripts/behavior_node_dummy.py
@@ -1,0 +1,26 @@
+# This node is a dummy node that will be used to test the behavior tree
+# It emulates voice commands and publishes to the behavior node
+# It emulates vision information and publishes to the behavior node
+
+import rospy
+from std_msgs.msg import String
+
+
+class DummyBehaviorNode:
+    def __init__(self):
+        self.command_pub = rospy.Publisher('speech_node/command', String, queue_size=10)
+        self.recognized_notes_pub = rospy.Publisher('vision_node/recognized_notes', String, queue_size=10)
+        self.main_loop()
+
+    def main_loop(self):
+        """
+        Wait for keyboard input, concatenate the input into a string, and send it to the server once the user presses enter.
+        """
+        while not rospy.is_shutdown():
+            # get input from user and create goal
+            input_str = input("Enter a command (e.g. 'marimbabot read'): ")
+            recognized_notes_str = input("Enter recognized notes (default 'c4 d4 e4'): ", )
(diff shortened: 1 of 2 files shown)


marimbabot_behavior/scripts/behavior_node_dummy.py
```diff
@@ -0,0 +1,27 @@
...
+        self.main_loop()
+
+    def main_loop(self):
+        """
+        Wait for keyboard input, concatenate the input into a string, and send it to the server once the user presses enter.
+        """
+        while not rospy.is_shutdown():
+            # get input from user and create goal
+            input_str = input("Enter a command (e.g. 'marimbabot read'): ")
+            recognized_notes_str = input("Enter recognized notes (default 'c4 d4 e4'): ", )
```
yunlong-wang-cn, line 23: thanks, What I just need it's actually this one. It's pretty clear to me.

yunlong-wang-cn (approved):

marimbabot_behavior/scripts/behavior_node_dummy.py
yunlong-wang-cn: I mean, I just need to know, which music seq or which data type t0 send.

Flova, line 23: New vision input and commands are coupled together. You would not want to update the vision data if you only want to send a `marimbabot slower`. I personally do not really see the benefits over a simple `rostopic pub ...` in the current state. Maybe if we would add numbered choices for commands, a simple gui or something like that.

Flova (changes requested):

marimbabot_behavior/scripts/behavior_node_dummy.py
Juphex: @yunlong-wang-cn

Juphex: @yunlong-wang-cn 
If we don't merge this branch you could still use this locally if it is handy to you  😅

Juphex:
I close this for now until a response from @yunlong-wang-cn

yunlong-wang-cn:
@Juphex which response do you means? I think I already give approvement of this PR.

Issue #216, opened 2023-08-10
Project Overview Documentation and Block diagram


ateRstones:
https://github.com/UHHRobotics22-23/marimbabot/tree/doc/block_diagram

Juphex:
I assume this can be closed since we merged the PR @ateRstones

PR #246, opened 2023-08-23
Project Overview and Block Diagram

ateRstones:
Regarding #216


berkgungor (commented):
Overall, diagram looks pretty good. However, at the end we can show some "play - result" box after planning section. We can maybe combine two arrows from planning and hardware sections and point them to a "play" part as end box. Because mallet hardware control is not the end of our motion but a part of the motion in practice. So I think diagram should not end with hardware node. @Flova what do you think ?

Flova:
> Overall, diagram looks pretty good. However, at the end we can show some "play - result" box after planning section. We can maybe combine two arrows from planning and hardware sections and point them to a "play" part as end box. Because mallet hardware control is not the end of our motion but a part of the motion in practice. So I think diagram should not end with hardware node. @Flova what do you think ?

The question would be if we want to include external nodes and if so how deep. Also the motion execution via moveit is only implicitly done with the ros interfaces under the hood.

Flova (approved):

Juphex (commented):
Maybe we re-name it to something like Node-Diagram/Node-Block Diagram/ROS Diagram?

PR #254, opened 2023-09-06
add flag for creating data without rendering images

Juphex:
also implement logic for when paths already exist, so it does not override the .ly/.txt files


Flova:
We can still merge this, but I probably wont touch the data gen in this stage anymore so idk I don't want to break it.

Juphex:
> We can still merge this, but I probably wont touch the data gen in this stage anymore so idk I don't want to break it.

Let us close it then.

PR #255, opened 2023-09-06
add logic for the repeat-parsing issue

Juphex:
I added a function to do that. 
That should be enough, right?

For testing I used the input string:
`r"\repeat volta 2 \key a \minor e'8 g'8 b'8 <e'' b''>8 ds'2 e'8 <d'' bs''>8 f'8 b'8 c''4 a'8 g'8 f'1"`

I also removed unused code/imports


Flova:
I think it also needs to be added to the visualization

Flova:
Otherwise lgtm

Juphex:
sure I will take a look at it

Flova (approved):

PR #257, opened 2023-09-14
 Adjust Timing information that is returned by planning

Juphex:
This branch restructures how the HitSequence message for the audio is generated.
As described in the original issue, this logic now should calculate a ground truth of the desired note sequence based on the absolute time when the first note was hit/played by the robot.

On the side of the audio:
The audio is not affected directly.
It is just important, that it gets this information after the robot executed the action, because of the current logic. Correct me if I am wrong @yunlong-wang-cn

### DIFF
--- a/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
+++ b/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
@@ -10,7 +10,8 @@ from std_msgs.msg import String
 
 from marimbabot_behavior.interpreter import read_notes
 from marimbabot_msgs.msg import (Command, HitSequence, HitSequenceAction,
-                                 LilypondAudioAction, LilypondAudioGoal)
+                                 HitSequenceElement, LilypondAudioAction,
+                                 LilypondAudioGoal)
 
 
 class ActionDecider:
@@ -171,13 +172,35 @@ class ActionDecider:
     """
     def planning_feedback_cb(self, feedback_msg):
         rospy.logdebug(f"Feedback from planning action server: {feedback_msg}")
-        # send it to audio '/audio/hit_sequence' topic
+        # the feedback message is a ROS Time value that contain the absolute time when the first note of the sequence was hit
+        # Calculate the relative time for each note in order create a HitSequence message
+        absolute_start_time: rospy.Time = feedback_msg.first_note_hit_time
         hit_sequence_msg = HitSequence()
         hit_sequence_msg.header.stamp = rospy.Time.now()
         hit_sequence_msg.sequence_id = self.sequence_id_counter
-        hit_sequence_msg.hit_sequence_elements = feedback_msg
+        
+        # create a list of HitSequenceElement messages
+        hit_sequence_elements : list(HitSequenceElement) = []
 
+        # iterate over the notes in the self.hit_sequence and create a HitSequenceElement message for each note
+        for elem in self.hit_sequence.notes:
+            hit_sequence_element = HitSequenceElement()
+            hit_sequence_element.tone_name = elem.tone_name
+            hit_sequence_element.octave = elem.octave
+            # add the relative start time to the absolute start time
+            hit_sequence_element.start_time = absolute_start_time + rospy.Duration(elem.start_time)
+            hit_sequence_element.tone_duration = elem.tone_duration
+            hit_sequence_element.loudness = elem.loudness
+
+            hit_sequence_elements.append(hit_sequence_element)
+
+        # add the list of HitSequenceElement messages to the HitSequence message
+        hit_sequence_msg.hit_sequence_elements = hit_sequence_elements
+
+        # increment the sequence_id counter
         self.sequence_id_counter += 1
+
+        # send it to audio '/audio/hit_sequence' topic
         self.hit_sequence_pub.publish(hit_sequence_msg)
         
     # communicates with the planning action server to play the hit sequence on the marimba
--- a/marimbabot_planning/include/marimbabot_planning/planning.h
+++ b/marimbabot_planning/include/marimbabot_planning/planning.h
@@ -17,7 +17,6 @@
 #include <tf2_ros/transform_listener.h>
 #include <tf2/LinearMath/Quaternion.h>
 #include <trajectory_msgs/JointTrajectoryPoint.h>
-#include <tuple>
 
 namespace marimbabot_planning
 {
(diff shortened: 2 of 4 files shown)


Juphex:
@Flova let me know whether I made the calculations for the first hit accurately

Flova:
Ping @Juphex

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -80,16 +82,40 @@ def change_tempo(self, faster=True, value=20):
...
         hit_sequence_msg = HitSequence()
         hit_sequence_msg.header.stamp = rospy.Time.now()
         hit_sequence_msg.sequence_id = self.sequence_id_counter
-        hit_sequence_msg.hit_sequence_elements = feedback_msg
+        
+        # create a list of HitSequenceElement messages
+        hit_sequence_elements : list(HitSequenceElement) = []
+
+        # iterate over the notes in the self.hit_sequence and create a HitSequenceElement message for each note
+        for i in range(len(self.hit_sequence.notes)):
```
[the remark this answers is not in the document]
Juphex: yes

marimbabot_planning/src/planning.cpp
```diff
@@ -215,10 +215,18 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::hit_note(
...
     auto plan = concatinated_plan({approach_plan, down_plan, retreat_plan});
+
+    // get the time when the plan when down_plan ends
+    ros::Time down_plan_end_time = start_time + down_stroke_duration;
+
+    // check whether approach_time_to_first_note_hit is set
+    if(this->approach_time_to_first_note_hit == ros::Duration(0.0))
+    {
+        // set approach_time_to_first_note_hit to the time 
+        this->approach_time_to_first_note_hit = down_plan.trajectory_.joint_trajectory.points.back().time_from_start + \
```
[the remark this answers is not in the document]
Juphex: I don't think I really get what you say.
Will "calling the function" you mean "hit_note", right?
After the first time the function was called from "hit_notes"?

I mean here:
```

moveit::planning_interface::MoveGroupInterface::Plan Planning::hit_notes(
    const moveit_msgs::RobotState& start_state,
    std::vector<CartesianHitSequenceElement> points)
{
    // Assert that there is at least one hit_point with an nice error message
    assert(points.size() > 0 && "There must be at least one hit_point");

    // Calculate hit trajectory
    auto hit_plan = hit_note(
        start_state,
        points.front()
        );

    // Call hit_points recursively for all remaining hit_points
    if(points.size() > 1)
    {
        auto remaining_hit_plan = hit_notes(
            get_robot_state_after_plan(hit_plan),
            std::vector<CartesianHitSequenceElement>(points.begin() + 1, points.end())
            );
        hit_plan = concatinated_plan({hit_plan, remaining_hit_plan});
    }

    return hit_plan;
}
```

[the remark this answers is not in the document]
Flova: Both would return the timing. Hit note would return the `hit_plan` as well as the time to the hit and `hit_notes` would return the time to hit of the first note in the sequence.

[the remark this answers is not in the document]
Juphex: I see your point.
Due to the recursive nature, I do not need an if condition anymore. When I set the time every time, I override it per se, and the last override happens for the first note, i.e. the first function call before .

Juphex (commented):
@yunlong-wang-cn and I talked about this and it looks fine.
Smaller things like hierarchical launch files and sorting imports are good to have.
It seems to be working. Yunlong will also add some README documentation on the launch files.

First, we should merge this branch locally on the robot's computer and try to integrate it before merging it in the origin/main.

Juphex (commented):
I rebased my branch and messed the commits up.
I force pushed my changes again @Flova

Flova:
Instead of using a class variable I would prefer using the return value to pass approach time. This is less prone to side effects. You also need to consider that the approach time not the time until the first note is played. So it is slightly off. The time until the note is played consists of the approach time + the down stroke duration.

Juphex (commented):
remove class variable

marimbabot_planning/include/marimbabot_planning/planning.h
```diff
@@ -17,6 +17,7 @@
 #include <tf2_ros/transform_listener.h>
 #include <tf2/LinearMath/Quaternion.h>
 #include <trajectory_msgs/JointTrajectoryPoint.h>
+#include <tuple>
```
Flova, line 20: I would have used a `std::pair` instead.

marimbabot_planning/src/planning.cpp
```diff
@@ -263,35 +266,36 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::hit_note(
...
     assert(points.size() > 0 && "There must be at least one hit_point");
 
     std::vector<ros::Duration> note_hit_times;
+
     // Calculate hit trajectory
-    auto hit_plan = hit_note(
-        start_state,
-        points.front()
-        );
+    std::tuple<moveit::planning_interface::MoveGroupInterface::Plan, ros::Duration> result = hit_note(start_state, points.front());
```
Flova, line 281: Modern C++ supports unpacking. Sadly we don't use C++17, which supports python style unpacking, but we can still simplify the code to:

```c++
moveit::planning_interface::MoveGroupInterface::Plan hit_plan;
ros::Duration approach_time_to_first_note_hit;
std::tie(hit_plan, approach_time_to_first_note_hit) = hit_note(start_state, points.front());
```

in C++17 this would be possible:

```C++
auto [hit_plan, approach_time_to_first_note_hit] = hit_note(start_state, points.front());
```

marimbabot_planning/src/planning.cpp
```diff
@@ -263,35 +266,36 @@ moveit::planning_interface::MoveGroupInterface::Plan Planning::hit_note(
...
+    moveit::planning_interface::MoveGroupInterface::Plan hit_plan = std::get<0>(result);
+    ros::Duration approach_time_to_first_note_hit = std::get<1>(result);
 
     // Call hit_points recursively for all remaining hit_points
     if(points.size() > 1)
     {
-        auto remaining_hit_plan = hit_notes(
+        auto remaining_hit_plan = std::get<0>(hit_notes(
```
Flova, line 288: If you use a tuple you could access it with `.first`

marimbabot_planning/src/planning.cpp
```diff
@@ -329,7 +332,9 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
         auto hits_relative_with_chords = apply_chords(hits_relative);
 
         // Define hit plan
-        auto hit_plan = hit_notes(start_state, hits_relative_with_chords);
+        std::tuple<moveit::planning_interface::MoveGroupInterface::Plan, ros::Duration> result =hit_notes(start_state, hits_relative_with_chords);
```
Flova, line 335: See the unpacking from above

marimbabot_planning/src/planning.cpp
Juphex: yes I stumpled accross that as well

PR #259, opened 2023-09-21
add explanation for data collection

Juphex:


marimbabot_vision/scripts/real_world_data_collection.py
```diff
@@ -19,8 +19,7 @@
     if not os.path.exists(output_file + "_0.png"):
         cv2.imshow("Data collection", 
             cv2.rotate(
-                cv2.imread(input_file), 
-                cv2.ROTATE_90_CLOCKWISE))
+                cv2.imread(input_file)))
```
Juphex, line 22: This is not needed anymore because the data generation itself rotates it now.

Flova (approved):

PR #260, opened 2023-09-25
Simple double mallet chord playing

### LINKED ISSUE #215: MVP Two Mallet Planning
Assume we always hit two notes at the same time and adjust the current planning for this

Flova:
This PR detects chords and plays them with two mallets.

Simpler and more robust version of #243 

Closes #215


Juphex:
Are there any additional restrictions we need to keep in mind regarding chords?
Like distance-wise

Flova:
> Are there any additional restrictions we need to keep in mind regarding chords? Like distance-wise

Chords that are not playable by the robot throw an ik error that is communicated to the user if he wants to play a melody that is not feasible. I don't think we need to handle this explicitly as we don't have an alternative way to play the chords.

Flova:
The PR is tested and ready for review btw.

marimbabot_description/urdf/marimbabot_ur5.urdf.xacro
```diff
@@ -130,7 +130,7 @@
   <joint name="world_marimba_joint" type="fixed">
     <parent link="world" />
     <child link = "marimba" />
-    <origin xyz="1.25 0.8 0.80" rpy="0.0 0.0 ${pi}" />
```
Juphex, line 133: What was the problem here? :)

marimbabot_planning/src/planning.cpp
```diff
@@ -60,68 +60,94 @@ void Planning::go_to_home_position()
  *goal
  * @param  start_state
  * @param  goal_point
```
Juphex, line 62: please rm this line and add
@param left_mallet_goal_point
@param right_mallet_goal_poin

marimbabot_planning/src/planning.cpp
```diff
@@ -274,8 +306,8 @@ void Planning::action_server_callback(const marimbabot_msgs::HitSequenceGoalCons
 
     try {
         // Set the max velocity and acceleration scaling factors
-        move_group_interface_.setMaxVelocityScalingFactor(0.5);
-        move_group_interface_.setMaxAccelerationScalingFactor(0.5);
+        move_group_interface_.setMaxVelocityScalingFactor(0.95);
```
Juphex, line 309: This factor does not multiply with the "original"/true speed of the robot but the restrictions that is was set to in the config file, right?

marimbabot_planning/src/utils.cpp
```diff
@@ -242,4 +249,33 @@ std::vector<CartesianHitSequenceElement> hit_sequence_absolute_to_relative(
     return hit_sequence_relative;
 }
 
+
+std::vector<CartesianHitSequenceElement> apply_chords(std::vector<CartesianHitSequenceElement> hits_relative)
```
Juphex, line 253: Could you add documentation here?

Juphex: And also for the other function that are introduced/have no documentation`? :)

marimbabot_ur5_flex_double_moveit_config/config/marimbabot_ur5.srdf
```diff
@@ -108,7 +108,7 @@
     <!--END EFFECTOR: Purpose: Represent information about an end effector.-->
     <end_effector name="s_model_tool0" parent_link="s_model_tool0" group="gripper" parent_group="arm"/>
     <end_effector name="mallet_head_1" parent_link="mallet_head_1" group="mallet_head_1_ee" parent_group="arm_mallets"/>
-    <end_effector name="mallet_head_2" parent_link="mallet_head_2" group="mallet_head_2_ee" parent_group="arm_mallets"/>
+    <!--<end_effector name="mallet_head_2" parent_link="mallet_head_2" group="mallet_head_2_ee" parent_group="arm_mallets"/>-->
```
Juphex, line 111: Why is this commented?

Juphex (approved):
I took my time to review the code and approve it because I am not able to provide a fast second feedback.

**Please see my comments and adjust what I mentioned there before merging :)**


It would be nice, if you have mentioned that I need to update the bio_ik repository somewhere. I wasted some time there :sweat_smile: 

I had trouble watching 3 different arms moving in my RVIZ but everything works fine. I used the sound_pay.py.

It is definitely worth it to add some documentation (README). I see that you didn't do it. I created a different issue for this. You may tackle it here or anyone else could grab it https://github.com/UHHRobotics22-23/marimbabot/issues/263.


**BTW, is there any way how to fix that my arm starts inside the wall? **
Is it just me?

image

marimbabot_planning/src/planning.cpp
Flova: Exactly. If set to `1.0` moveit returns the fastest plan for the given planner and constraints. You can lower the speed of the robot by decreasing it. But if is different to the robots velocity and acceleration limits which are statically defined. Idk why it was so low tbh.. I think someone might decreased it because the robot failed to follow the trajectory, but this was most likely a different issue where the interpolation returned Nan for trajectories with a length of 1.

marimbabot_planning/src/utils.cpp
Flova: There is a doc string in the header file, that is shown if the IDE is configured properly, but I can copy it as we did with the other ones.

marimbabot_description/urdf/marimbabot_ur5.urdf.xacro
Flova: The marimba was too far away, so we where not able to reach certain black keys. This was solved in a hacky way by using approximate ik solutions, but it doesn't hurt to pull it a little closer and use the exact solutions.

marimbabot_planning/src/planning.cpp
Flova: Good point

marimbabot_ur5_flex_double_moveit_config/config/marimbabot_ur5.srdf
Flova: Moveit throws an error if the robot has two end effectors. Reducing it to one does not change the planning code, but it reduces spam in the terminal. As we set the mallet position goals together with a bunch of other constraints via the ik options and not the normal API it doesn't change anything for us. It does change the behavior in rviz tho. You are no longer able to directly interact with both mallets in the 3d view.

Flova:
@Juphex I read your feedback and will address it. I also added some comments to your feedback.

You can add the `RobotModel` plugin to rviz and hide the moveit planning one to only view the current robot state.

marimbabot_planning/src/utils.cpp
Juphex: If it is already there then nvm

berkgungor:
I am on wall collision issue.  + #263

PR #265, opened 2023-09-28
Data/add scale notes for data

cvhex:


Flova:
Can you give a more detailed description what this PR does?

Juphex:
I do not think we should merge this. It is just a simple adaptation for collecting keys/bpm only and has no further usage in our main branch. 😄

PR #268, opened 2023-09-28
Fix double mallet start point collision

berkgungor:
- updated initial pose for demo.launch to avoid collision

### DIFF
--- a/marimbabot_ur5_flex_double_moveit_config/config/marimbabot_ur5.srdf
+++ b/marimbabot_ur5_flex_double_moveit_config/config/marimbabot_ur5.srdf
@@ -96,6 +96,15 @@
         <joint name="ur5_wrist_2_joint" value="1.6173"/>
         <joint name="ur5_wrist_3_joint" value="0"/>
     </group_state>
+    <group_state name="marimbabot_home" group="arm_mallets">
+        <joint name="ur5_elbow_joint" value="-2.1254"/>
+        <joint name="ur5_shoulder_lift_joint" value="-1.947"/>
+        <joint name="ur5_shoulder_pan_joint" value="-1.3375"/>
+        <joint name="ur5_wrist_1_joint" value="-0.491"/>
+        <joint name="ur5_wrist_2_joint" value="1.6173"/>
+        <joint name="ur5_wrist_3_joint" value="0"/>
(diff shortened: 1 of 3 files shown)


marimbabot_ur5_flex_double_moveit_config/config/marimbabot_ur5.srdf
```diff
@@ -96,15 +96,6 @@
         <joint name="ur5_wrist_2_joint" value="1.6173"/>
         <joint name="ur5_wrist_3_joint" value="0"/>
     </group_state>
-    <group_state name="marimbabot_home" group="arm_mallets">
```
Flova, line 99: Why did you delete the home position? It is utilized at multiple points in the planning pipeline?

berkgungor: two marimbabot_home was defind and looked like a duplicate. but just saw that one line is different "mallet_finger" in second marimbabot_home group. I think we should delete the first defined group.

Flova: We can keep both as they are configurations for different groups. One is for "arm_mallets" and includes a state for the `mallet_finger` and one is only for the `arm` (the old single mallet setup). We can keep both as they do not result in any problems and define a homing position for both groups.

Flova: I would revert [8997cc3](https://github.com/UHHRobotics22-23/marimbabot/pull/268/commits/8997cc3b00e4c0c0dedd05830bb06b81417c6b18)

marimbabot_ur5_moveit_config/config/fake_controllers.yaml
```diff
@@ -26,4 +26,4 @@ initial:  # Define initial robot poses per group
   - group: gripper
     pose: closed
   - group: arm
-    pose: home
\ No newline at end of file
+    pose: marimbabot_home
```
Juphex, line 29: looks good

Juphex (commented):
works but please see https://github.com/UHHRobotics22-23/marimbabot/pull/268/files#r1341241545

Flova (approved):

PR #273, opened 2023-10-03
translate repeat to parseable format

scaredycode:
Checks the recognized note sequence for repeats, then modifies the internal note sequence to the full note string (without repeat sign)

Result:
recognized notes: \repeat volta 2 \tempo 4=60 \key f \major c'4 d'4 e'4 f'4 g'4 a'4 b'4 c''4
[INFO] [1696327250.492832]: updated notes: \tempo 4=60 \key f \major c'4 d'4 e'4 f'4 g'4 a'4 b'4 c''4 c'4 d'4 e'4 f'4 g'4 a'4 b'4 c''4
[INFO] [1696327251.099850]: created midi file test_midi.midi, success: True

Caution: Only supported if "\repeat volta 2" is **at the beginning** of the note sequence and the whole piece is to be repeated (**no partial repeats**).


marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -49,6 +49,19 @@ def __init__(self):
...
+        if '\\repeat volta 2' in self.note_sequence:
+            # get the index of the first note
+            note_sequence_list = self.note_sequence.split(' ')
+            first_note_index = None
+            for i, x in enumerate(note_sequence_list):
+                if re.match(r'[a-g]\'*[0-9]+(\>)?(\.)?', x):
+                    first_note_index = i
+                    break
+            self.note_sequence = ' '.join(note_sequence_list[3:] + note_sequence_list[first_note_index:])
+            rospy.loginfo(f"updated notes: {self.note_sequence}")
```
Juphex, line 63: I think it would be better to use logdebug here?

scaredycode: I use loginfo a lot in the behavior.. Whenever I modify the lilypond sequence, I like to immediately print the result. Should I replace it with logdebug wherever I print the (modified) note sequence (genuine question)?

Flova: Normally, applications should be "as silent as possible" if everything is fine, as otherwise the terminal output grows in size very quickly for large applications. You can explicitly set certain nodes to a default loglevel or view their logs in rqt if needed.

marimbabot_behavior/launch/test_behavior.launch
```diff
@@ -1,7 +1,6 @@
 <launch>
 <node name="behavior_node" pkg="marimbabot_behavior" type="behavior_node.py" output="screen"/>
 <node name="audio_from_lilypond" pkg="marimbabot_audio" type="audio_from_lilypond.py"/>
-<include file="$(find marimbabot_audio)/launch/midi.launch"/>
```
Flova, line 4: Why did you remove this?

Flova (approved):

marimbabot_behavior/launch/test_behavior.launch
scaredycode: This launchfile no longer exists. For testing, I needed to comment out this line. Preview and TTS still function.

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
scaredycode: I see. I will change it then :+1:

marimbabot_behavior/launch/test_behavior.launch
Flova: Ah okay!

Juphex (commented):
Could you also use the existing function 
create_staff_from_notes
or 
create_lilypond_file ?
Is it not possible with a LilypondFile object?

Juphex (changes requested):
function "create_lilypond_file" in interpreter.py might be obsolete / could be removed

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -279,13 +290,14 @@ def callback_command(self, command_msg):
...
+            rospy.logdebug('reading notes')
             # update the note_sequence variable with the latest note sequence from vision_node/recognized_notes to signal that notes have been read
             try:
                 # wait for the vision node to publish a recognized note sequence
                 self.note_sequence = rospy.wait_for_message('vision_node/recognized_notes', String, timeout=5).data 
                 rospy.loginfo(f"recognized notes: {self.note_sequence}")
-                self.response_pub.publish('Notess recognized.')
+                self.response_pub.publish('Notes recognized.')
+                self.check_for_repeat()
                 self.update_hit_sequence()
```
Juphex, line 301: in this function call below, we also check for repeats

scaredycode:
> Could you also use the existing function create_staff_from_notes or create_lilypond_file (both in interpreter)? Is it not possible with a LilypondFile object?
> 
> Because now, if you adjust the self.note_sequence these functions get obsolete because they also check for a "\repeat" but you remove them. If I am not mistaking

My reasoning was, that I also need a parseable string for the preview, which would run into the same issue if I only apply changes in the interpreter.

marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
Juphex: You are right, but here
first 
`self.check_for_repeat()` is called which operates on `self.note_sequence`.
Then, ` self.update_hit_sequence()` is called. In this function, `self.hit_sequence = read_notes(self.note_sequence)` sends the new `self.note_sequence` to `read_notes` in which at some part  `create_staff_from_notes` is called. In this function, there is also a check for a repeat.
Therefore, we check whether there is a repeat in `self.note_sequence` twice.

Could you adjust that, so we just do it in the behavior_node? :)

scaredycode: I removed all repeat logic from the interpreter. This should do the trick :+1:

Juphex: Nice, but do you agree with what I said? 
Because that was my thought process

scaredycode: Yeah I do agree. My changes in the behavior would cause the interpreter to skip the second repeat check every time. You are right that these lines should be removed. Good catch

scaredycode:
> > > Could you also use the existing function create_staff_from_notes or create_lilypond_file (both in interpreter)? Is it not possible with a LilypondFile object?
> > > Because now, if you adjust the self.note_sequence these functions get obsolete because they also check for a "\repeat" but you remove them. If I am not mistaking
> > 
> > 
> > My reasoning was, that I also need a parseable string for the preview, which would run into the same issue if I only apply changes in the interpreter.
> 
> I thought about this here [#273 (comment)](https://github.com/UHHRobotics22-23/marimbabot/pull/273#discussion_r1344047809). Immediately after your `self.check_for_repeat()` function `self.update_hit_sequence()` is called. In both logics, a check for repeats is performed. Therefore, I assume the behavior preview is not affected when you remove the second check. Additionally, the second check is only triggered once by the behavior node.
> 
> ```
>     def callback_command(self, command_msg):
>         # command = command_msg.command.lower()
>         rospy.loginfo(f"received command: {command_msg}")
> 
>         # read notes on the whiteboard
>         if command_msg.behavior == "read":
>             rospy.logdebug('reading notes')
>             # update the note_sequence variable with the latest note sequence from vision_node/recognized_notes to signal that notes have been read
>             try:
>                 # wait for the vision node to publish a recognized note sequence
>                 self.note_sequence = rospy.wait_for_message('vision_node/recognized_notes', String, timeout=5).data 
>                 rospy.loginfo(f"recognized notes: {self.note_sequence}")
>                 self.response_pub.publish('Notes recognized.')
>                 self.check_for_repeat()
>                 self.update_hit_sequence()
>             except rospy.ROSException:
>                 rospy.logwarn('No notes recognized.')
>                 self.response_pub.publish('No notes recognized.')
> ```
> 
> Am I correct?

Yes you are absolutely right. The preview is not at all affected anything that happens in the interpreter, which is only used to calculate the hit sequence for the motion planning node. The first check in the behavior should fix both preview and planning messages.

PR #275, opened 2023-10-04
fix marcato check

scaredycode:
Marcato was causing problems if written behind a dynamic symbol. The fix now adds an additional check so that the order does not matter anymore.
Tested with 10 samples generated using the latest data generation version (+ articulations, +dynamics).

Also fixes a placement error with dotted chords and a false "unregognized" message for repeats

### DIFF
--- a/marimbabot_vision/scripts/generate_data.py
+++ b/marimbabot_vision/scripts/generate_data.py
@@ -18,9 +18,9 @@ NUM_SAMPLES = 10000
 NUM_WORKER = 24
 OUTPUT_DIR = "data"
 MIN_DURATION = 8 # 1/8th note
-INCLUDE_DYNAMICS = False
+INCLUDE_DYNAMICS = True
 INCLUDE_SLURS = False
-INCLUDE_ARTICULATIONS = False
+INCLUDE_ARTICULATIONS = True
 INCLUDE_SCALES = True
 INCLUDE_REPEATS = True
 INCLUDE_CHORDS = True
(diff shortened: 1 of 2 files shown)


marimbabot_vision/scripts/generate_data.py
```diff
@@ -18,9 +18,9 @@
 NUM_WORKER = 24
 OUTPUT_DIR = "data"
 MIN_DURATION = 8 # 1/8th note
-INCLUDE_DYNAMICS = False
+INCLUDE_DYNAMICS = True
 INCLUDE_SLURS = False
-INCLUDE_ARTICULATIONS = False
+INCLUDE_ARTICULATIONS = True
```
Juphex, line 23: Why did you change the default parameters?

marimbabot_vision/scripts/generate_data.py
```diff
@@ -203,14 +203,14 @@ def generate_piece(self,num_bars=3,):
             self.add_tempo(voice_1)
         if self.slurs:
             self.add_slurs(voice_1)
+        if self.include_dynamics:
```
Juphex, line 206: > Also fixes a placement error with dotted chords and a false "unregognized" message for repeats

Is this here?

marimbabot_vision/scripts/generate_data.py
Flova: Because we agreed on dynamics earlier, but they were off, so we need to regenerate all the data -.- Same for the articulations. They were reduced to only marcato, but also deactivated.

Flova (approved):

marimbabot_vision/scripts/generate_data.py
scaredycode: Request from Flo. The default should be the state we want to use in the end. So with dynamics and articulations (marcato)

marimbabot_vision/scripts/generate_data.py
scaredycode: It was an attempt at changing the order of dynamics and articulations in the data generation (as this was causing issues in the handwritten data generation, because in a previous version, the order was switched). But it did not work unfortunately. I commited before I noticed the issue still occurs. I can switch it back if you'd like

PR #280, opened 2023-10-05
Fix initial volume assignment

scaredycode:
Fixes issue with non-updated note sequences after the louder/softer command was triggered.
Now, a default volume (\mf) is defined for the first note BEFORE any volume changes are performed.
The default volume assignment will NOT overwrite the volume of the first note (if given), but it will add the dynamic symbol if there is none.


Flova (approved):
Merge if tested

Juphex (approved):

PR #281, opened 2023-10-05
Add low frames for sharp notes

iibrahimli:
Add `bar_*_low` frames for sharp notes (top row), which are potentially faster to reach than the regular ones in the center of the bar. The sharp notes now have an additional frame, e.g. `bar_cis4` has a corresponding `bar_cis4_low`. The relevant part is shown in the screenshot below.

The frames are added by creating a dummy link of the same name, which has a 0-sized box as its visual geometry (it's required to have _something_ there) and no collision geometry.

The frame is not precisely at the tip of the bar, but rather somewhere between the tip and the mount to get the highest sound volume (tip would be easy to miss, and mount point would damp the sound).

image

### DIFF
--- a/marimbabot_description/urdf/marimba.urdf.xacro
+++ b/marimbabot_description/urdf/marimba.urdf.xacro
@@ -16,11 +16,11 @@ From the point of view of the marimbist:
     z: points towards the ceiling, according to right hand rule
 
 Changelog:
+    2023.01.04 First version created
+    2023.01.19 Row of keys working
+    2023.01.26 Complete model v1
     2023.03.02 Named and correctly positioned bar frames
                Added side boxes
-    2023.01.26 Complete model v1
-    2023.01.19 Row of keys working
-    2023.01.04 First version created
     2023.06.14 Made side boxed width and height smaller
 -->
 
--- a/marimbabot_description/urdf/marimba.urdf.xacro
+++ b/marimbabot_description/urdf/marimba.urdf.xacro
@@ -101,7 +101,7 @@ Changelog:
 
 
 <!-- The bar -->
-<xacro:macro name="bar" params="note parent id pos length" >
+<xacro:macro name="bar" params="note parent id pos length add_low_frame:=0" >
     <gazebo reference="bar_${note}" >
         <turnGravityOff>false</turnGravityOff>
         <selfCollide>true</selfCollide>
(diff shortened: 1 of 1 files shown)


Flova:
Otherwise LGTM

marimbabot_description/urdf/marimba.urdf.xacro
```diff
@@ -152,18 +152,36 @@ Changelog:
...
+            <origin xyz="${length / 2 - 0.02} 0 0" rpy="0 0 0" />
+        </xacro:tams_fixedjoint>
+    </xacro:if>
 </xacro:macro>
 
 
 <!-- Wrapper for a row of bars, computes the position and length based on fixed measurements -->
 <xacro:macro name="bar_in_row" params="note parent id left_x left_y left_z length_diff bar_length_left n_bars slope:=0 mirror:=1" >
     <!-- mirror controls alignment of bars in row: -1 (right align) or 1 (left align) -->
+    <!-- mirror is set to -1 on the top row (sharp notes), and therefore is also used to control the low frame -->
```
Juphex, line 176: Is the value of "mirror" to be changed inside this line of code?

iibrahimli: Works correctly as is

marimbabot_description/urdf/marimba.urdf.xacro
```diff
@@ -101,7 +101,7 @@ Changelog:
 
 
 <!-- The bar -->
-<xacro:macro name="bar" params="note parent id pos length" >
```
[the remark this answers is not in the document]
iibrahimli: The low frames are active by default, if that's what you mean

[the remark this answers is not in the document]
iibrahimli: The default is used for the bottom row (no low frames there), otherwise is set to 1 [here](https://github.com/UHHRobotics22-23/marimbabot/pull/281/files#diff-20fb49e2c5b33e5c29b5f7cb39e71b2085e6906d119e054b22e71df8c53c70ceR184)

marimbabot_description/urdf/marimba.urdf.xacro
iibrahimli: the 1 above is the default (used in bottom row), otherwise this is set to -1 when calling the macro for the top row. Now this variable also controls whether the low frames are added or not

Juphex (approved):

PR #283, opened 2023-10-05
Use the lower notes for sharps

Flova:
PR #281 adds new hit points and this PR utilizes them in the motion.


Flova:
This is totally untested as of now and waits for #283

iibrahimli:
LG if either `ends_with` is implemented, or we switch to C++ 20 :D

Juphex (commented):
perhaps you could use any function from https://stackoverflow.com/questions/874134/find-out-if-string-ends-with-another-string-in-c

Flova:
Well the old CPP version strikes again...
We can use the boost ends_with tho. similar to the boost optional we are using.

berkgungor:
I know it is a different PR, but I also pushed changes that enable the preemption / canceling of a running action.

yunlong-wang-cn (approved):
lgtm

PR #285, opened 2023-10-07
Audio small patch

yunlong-wang-cn:
1. add one more threshold to filter out the no-speech noise or the background speech noise by the  `no_speech_preb` reported by Whisper.
2. remove the test log
3. add reference value for onset detection, to enhance the detection of music note from right zone of the marimba, also it has the function of the denoise.

### DIFF
--- a/marimbabot_audio/src/onset_detection.py
+++ b/marimbabot_audio/src/onset_detection.py
@@ -127,7 +127,7 @@ class OnsetDetection:
 			confidence threshold for note classification(crepe)
 		"""
 		# For onset detection
-		self.confidence_threshold = rospy.get_param("~confidence_threshold", 0.7)
+		self.confidence_threshold = rospy.get_param("~confidence_threshold")
 		self.amplitude_ref = rospy.get_param("~amplitude_ref", 10.0)
 		self.windows_for_classification = 0.1  # using 0.1 sec data after onset time for note classification
 		# preload model to not block the callback on first message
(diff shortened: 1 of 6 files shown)


marimbabot_bringup/launch/marimbabot.launch
```diff
@@ -1,7 +1,7 @@
 <launch>
   <include file="$(find marimbabot_vision)/launch/marimbabot.launch"/>
-  <include file="$(find marimbabot_audio)/launch/marimbabot.launch"/>
-  <include file="$(find marimbabot_planning)/launch/marimbabot.launch"/>
   <include file="$(find marimbabot_speech)/launch/marimbabot.launch"/>
   <include file="$(find marimbabot_behavior)/launch/marimbabot.launch"/>
+  <include file="$(find marimbabot_audio)/launch/marimbabot.launch"/>
```
Juphex, line 5: Why did you change the order?

yunlong-wang-cn: they though it's much intuitive.  I forgot who made this change，since they decide to do this,  I just put it here. But from my perspective its doesnt matter.

marimbabot_audio/src/onset_detection.py
```diff
@@ -127,7 +127,8 @@ def init_detection_config(self):
 			confidence threshold for note classification(crepe)
 		"""
 		# For onset detection
-		self.confidence_threshold = 0.7  # the threshold for note classification
+		self.confidence_threshold = rospy.get_param("~confidence_threshold", 0.7)
```
Flova, line 130: I would not include a default, as it might cause confusions when somebody changes that.

Flova (approved):

marimbabot_audio/src/onset_detection.py
yunlong-wang-cn: done

PR #296, opened 2023-10-12
fix volume assignment for chords + repeat from GUI

scaredycode:


Juphex (commented):
Is this the working code?

Flova (approved):

Flova:
Looks good, but we still need to test it.

yunlong-wang-cn (approved):
If you tested the re pattern, then I think it is good to go.

PR #302, opened 2023-10-16
Add decision tree to Behavior README

scaredycode:
added decision tree

### DIFF
--- a/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
+++ b/marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
@@ -319,22 +319,20 @@ class ActionDecider:
                 self.response_pub.publish('No notes recognized.')
 
         elif command_msg.behavior == "stop":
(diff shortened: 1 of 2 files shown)


marimbabot_behavior/src/marimbabot_behavior/behavior_node.py
```diff
@@ -369,24 +387,6 @@ def callback_command(self, command_msg):
                     rospy.logwarn('The audio preview is already playing.')
                     self.response_pub.publish('The audio preview is already playing.')
 
-            elif command_msg.behavior == "stop":
```
Juphex, line 372: Why did you change this? @scaredycode

scaredycode: To match the decision tree. It has no negative effects on the functionality. A positive effect is that we will no longer get the "No notes to play" Response for using the 'stop' command before a successful 'read'

